PR #23, opened 2025-01-09
fix: make labels take up less pie chart space

kuzmadom:


src/hooks/useShapes/pie/prepare-data.ts
```diff
@@ -93,165 +93,225 @@ export function preparePieData(args: Args): PreparedPieData[] {
...
+        }
 
-                    y = y < 0 ? y - labelHeight : y;
+        return data;
+    };
 
-                    if (shouldUseHtml) {
-                        x = x < 0 ? x - labelWidth : x;
-                    }
+    const prepareLabels = (args: {data: PreparedPieData; series: PreparedPieSeries[]}) => {
```
korvin89, line 110: could you rename `args` to avoid shadowing?

korvin89 (approved):

korvin89 (approved):

Issue #30, opened 2024-01-29
Plot lines, bands


kuzmadom:
Add ability to specify an array of colored bands or lines stretching across the entire area of the graph, marking the interval on the axis.

PR #45, opened 2025-03-04
feat: yAxis plotLines

kuzmadom:
related to https://github.com/gravity-ui/charts/issues/30

### DIFF
--- a/src/__stories__/__data__/bar-x/playground.ts
+++ b/src/__stories__/__data__/bar-x/playground.ts
@@ -48,6 +48,12 @@ function prepareData(): ChartData {
                         width: 2,
                         color: 'red',
                         dashStyle: 'Dash',
+                        layerPlacement: 'after',
+                    },
+                    {
+                        value: 200,
+                        width: 1,
+                        layerPlacement: 'before',
                     },
                 ],
             },
--- a/src/__stories__/__data__/bar-x/playground.ts
+++ b/src/__stories__/__data__/bar-x/playground.ts
@@ -42,6 +42,14 @@ function prepareData(): ChartData {
                 ticks: {
                     pixelInterval: 120,
                 },
+                plotLines: [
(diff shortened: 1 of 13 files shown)


src/__stories__/__data__/bar-x/playground.ts
```diff
@@ -42,6 +42,14 @@ function prepareData(): ChartData {
                 ticks: {
                     pixelInterval: 120,
                 },
+                plotLines: [
```
korvin89, line 45: Will we have a possibility to render plotLines above the main chart?
image

src/types/chart/axis.ts
```diff
@@ -80,6 +82,26 @@ export interface ChartAxis {
 
 export interface ChartXAxis extends ChartAxis {}
 
+export interface AxisPlotLine {
```
korvin89, line 85: Should `AxisPlotLine` include type of line `x` or `y`?

src/components/Axis/AxisY.tsx
```diff
@@ -215,6 +214,28 @@ export const AxisY = (props: Props) => {
...
             }
 
+            if (d.plotLines.length > 0) {
+                const plotLineClassName = b('plotLine');
+                axisSelection
+                    .selectAll(`.${plotLineClassName}`)
+                    .data(d.plotLines)
+                    .join('path')
+                    .attr('class', plotLineClassName)
+                    .attr('d', (d) => {
```
korvin89, line 224: Could u rename `d` from outer scope to remove eslint warnings?

src/types/chart/axis.ts
kuzmadom: The `AxisPlotLine` is contained inside the X or Y axis. The type is inherited from the parent

korvin89 (approved):

PR #59, opened 2025-04-30
feat: add radar series

kuzmadom:


korvin89:
Some thoughts about UX:
- should we show the tooltip so early?
https://github.com/user-attachments/assets/046bebed-08d1-4db8-b999-00d6aec9cf7a
- is it possible to accent an element with a hover in such way?
https://github.com/user-attachments/assets/768b7ec0-a89f-4d59-98c0-5a30cb2cebde

src/hooks/useShapes/radar/index.tsx
```diff
@@ -0,0 +1,177 @@
...
+        return () => {
+            dispatcher.on(eventName, null);
+        };
+    }, [dispatcher, preparedData, seriesOptions]);
+
+    const htmlElements = preparedData.map((d) => d.htmlLabels).flat();
+
+    return (
+        <React.Fragment>
+            <g ref={ref} className={b()} style={{zIndex: 9}} />
```
korvin89, line 173: Why `zIndex: 9`?

PR #62, opened 2025-05-12
feat: add dataLabels format options

kuzmadom:


src/hooks/useSeries/types.ts
```diff
@@ -150,6 +151,8 @@ export type PreparedBarXSeries = {
         allowOverlap: boolean;
         padding: number;
         html: boolean;
+        dateFormat?: string;
+        numberFormat?: FormatNumberOptions;
```
korvin89, line 155: This pattern looks like a separate type

korvin89 (approved):

PR #83, opened 2025-07-21
Add visible prop to axes

dvortsovs:


kuzmadom:
Снимок экрана 2025-07-21 в 18 19 16

The visibility of the axis should also affect the calculation of the size for the chart rendering area

src/types/chart/axis.ts
```diff
@@ -80,6 +80,7 @@ export interface ChartAxis {
     maxPadding?: number;
     /** An array of lines stretching across the plot area, marking a specific value */
     plotLines?: AxisPlotLine[];
+    visible?: boolean;
```
kuzmadom, line 83: Additional options such as this one should be accompanied by comments in the jsdoc format - documentation is based on them.

kuzmadom (changes requested):

src/types/chart/axis.ts
dvortsovs: fixed

src/hooks/useChartDimensions/utils.ts
```diff
@@ -16,6 +16,9 @@ export const getBoundsWidth = (args: {
 };
 
 export function getYAxisWidth(axis: PreparedAxis | undefined) {
+    if (!axis?.visible) {
+        return axis?.labels.margin || 0;
+    }
```
kuzmadom, line 21: It's the same here

kuzmadom (approved):

PR #95, opened 2025-07-30
fix: fix pie chart horizontal cropping in case of limited container width

korvin89:


src/hooks/useShapes/pie/prepare-data.ts
```diff
@@ -43,7 +43,13 @@ const getCenter = (
 
 export function preparePieData(args: Args): PreparedPieData[] {
     const {series: preparedSeries, boundsWidth, boundsHeight} = args;
-    const maxRadius = Math.min(boundsWidth, boundsHeight) / 2;
+    const isWidthLessThanHeight = boundsWidth < boundsHeight;
+    let maxRadius = Math.min(boundsWidth, boundsHeight) / 2;
+
+    if (isWidthLessThanHeight) {
```
kuzmadom, line 49: -

PR #108, opened 2025-08-12
fix(Pie): fix possible chart intersection by connectors & labels

korvin89:


korvin89:
Will be fixed here https://github.com/gravity-ui/charts/pull/109

PR #112, opened 2025-08-14
fix(Pie): fix html labels overflow

korvin89:


src/utils/chart/text.ts
```diff
@@ -128,19 +129,24 @@ export function getLabelsSize({
...
-
     const result = {maxHeight: 0, maxWidth: 0};
     let labelWrapper: HTMLElement | null;
+
     if (html) {
-        labelWrapper = container.append('div').style('position', 'absolute').node();
+        labelWrapper = container
+            .append('div')
+            .style('position', 'absolute')
+            .style('font-size', style?.fontSize || DEFAULT_DATALABELS_STYLE.fontSize)
```
kuzmadom, line 139: Are we sure that we use this method only for dataLabels? it may be worth passing the styles above, without using the default ones here.

PR #115, opened 2025-08-15
feat(Pie): add minRadius option

korvin89:


src/hooks/useShapes/pie/prepare-data.ts
```diff
@@ -49,7 +49,12 @@ export function preparePieData(args: Args): PreparedPieData[] {
         ? preparedSeries[0].states.hover.halo.size
         : 0;
     const maxRadius = Math.min(boundsWidth, boundsHeight) / 2 - haloSize;
-    const minRadius = maxRadius * 0.3;
+    const propsMinRadius =
+        calculateNumericProperty({
+            value: preparedSeries[0].minRadius,
+            base: maxRadius,
+        }) || 0;
+    const minRadius = Math.max(propsMinRadius, maxRadius * 0.3);
```
kuzmadom, line 57: If minRadius cannot be reduced by more than a third of the maximum, then maybe it's worth writing about it in typing's js-doc?
But I would expect it to be limited to just zero. And a third is just the default value for minRadius settings.

kuzmadom (approved):

PR #117, opened 2025-08-18
feat: add `legend.html` & `tooltip.html` options

korvin89:


src/__snapshots__/pie-series.visual.test.tsx-snapshots/Pie-series-Handle-html-legend-html-true-1-chromium-linux.png
kuzmadom, line 1: according to the screenshot, the legend is slightly cropped from the bottom, is this how it is intended?

kuzmadom (dismissed):

PR #119, opened 2025-08-21
fix: pie html dataLabels (color, textOverflow, etc)

kuzmadom:


playwright/components/ChartTestStory.tsx
```diff
@@ -1,21 +1,35 @@
...
-export const ChartTestStory = ({data, styles}: Props) => {
+export const ChartTestStory = ({data, styles, customShape}: Props) => {
     const storyStyles: React.CSSProperties = {
         height: 280,
         width: 400,
         display: 'inline-block',
         ...styles,
     };
 
+    if (customShape) {
```
kuzmadom, line 23: exclusively for playwright tests - it is not forwarded from the config

korvin89 (approved):

PR #137, opened 2025-08-26
fix(Pie): performance of the label rendering

kuzmadom:


src/utils/chart/text.ts
```diff
@@ -252,6 +252,23 @@ export function wrapText(args: {text: string; style?: BaseTextStyle; width: numb
     }, []);
 }
 
+function unescapeHtml(str: string) {
+    const entityMap = {
```
korvin89, line 256: We can move this map out of function

kuzmadom: fixed

korvin89 (approved):

PR #145, opened 2025-09-04
feat: add basic zoom feature

korvin89:
Added a feature to zoom data on axis related charts.

----

To enable it, you need to add the following setting:
```javascript
const data = {
  // other data
  chart: {
    zoom: {
      type: 'x',
    },
  },
}
```
There is no usual `enabled` flag here, since the type setting must be meaningful, which makes `type` setting mandatory.

----

How does it work:
- zooming occurs by selecting the area of the chart ([d3-brush](https://d3js.org/d3-brush) tool)
- at the end of the work, the selection boundaries are remembered
- the boundary values for the X and Y axes are determined based on the obtained boundaries (depending on the `zoom.type`)
- boundary values are used to filter `series.data`
- a new chart is being built based on the filtered data.

### DIFF
--- a/src/components/ChartInner/useChartInnerProps.ts
+++ b/src/components/ChartInner/useChartInnerProps.ts
@@ -34,12 +34,8 @@ export function useChartInnerProps(props: Props) {
     const prevHeight = usePrevious(height);
     const [zoomState, setZoomState] = React.useState<Partial<ZoomState>>({});
     const {chart, title, tooltip, colors} = useChartOptions({data});
-    // const {preparedSeries, preparedSeriesOptions} = useSeries({colors, seriesData, seriesOptions})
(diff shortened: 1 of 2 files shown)


src/__tests__/line-series.visual.test.tsx
korvin89, line 1: **Why did I correct this text?**
- For some reason, this instruction `await component.locator('.gcharts-line').hover();` stopped working without `force: true` setting
- After adding `force: true` setting, the tooltip was displayed half a pixel higher

Thus, considering that the test checks the display of a tooltip for a vertical line, I decided to change it so that it checks the display of a tooltip for both points. As a result, the position of the tooltip for both points became more stable, and the test began to check that two different tooltips were being shown.

kuzmadom (dismissed):

PR #148, opened 2025-09-10
feat: add `max` property to x-axis & y-axis

korvin89:


src/utils/chart/index.ts
```diff
@@ -132,6 +138,45 @@ export function getDefaultMaxXAxisValue(series: UnknownSeries[]) {
     return undefined;
 }
 
+export function getDefaultMinXAxisValue(series: UnknownSeries[]) {
+    if (series?.some((s) => CHART_SERIES_WITH_VOLUME_ON_X_AXIS.includes(s.type))) {
```
kuzmadom, line 142: There probably can't be any series with and without  axis in one chart, but it makes sense to check each of them, not the general set of series.

korvin89:
Will be added here https://github.com/gravity-ui/charts/pull/156

PR #157, opened 2025-09-18
feat: add `xAxis.max & yAxis.max properties`

korvin89:
Previous PR https://github.com/gravity-ui/charts/pull/148


src/utils/chart/index.ts
```diff
@@ -132,6 +138,45 @@ export function getDefaultMaxXAxisValue(series: UnknownSeries[]) {
     return undefined;
 }
 
+export function getDefaultMinXAxisValue(series: UnknownSeries[]) {
+    if (series?.some((s) => CHART_SERIES_WITH_VOLUME_ON_X_AXIS.includes(s.type))) {
+        return series.reduce((minValue, s) => {
+            // @ts-expect-error
```
kuzmadom, line 144: Is it something with the types? Why this comment?

src/utils/chart/index.ts
```diff
@@ -132,6 +138,45 @@ export function getDefaultMaxXAxisValue(series: UnknownSeries[]) {
...
+            const minYValue = s.data.reduce((res, d) => Math.min(res, get(d, 'x', 0)), 0);
```
kuzmadom, line 145: min**X**Value ?

kuzmadom (dismissed):

src/utils/chart/index.ts
kuzmadom: then is it worth checking and filtering?

korvin89: Will be fixed here https://github.com/gravity-ui/charts/issues/160

PR #179, opened 2025-10-02
feat: add `tooltip.totals` option

korvin89:


src/types/chart/tooltip.ts
```diff
@@ -114,4 +119,22 @@ export interface ChartTooltip<T = MeaningfulAny> {
...
+    /** Settings for totals block in tooltip */
+    totals?: {
+        /**
+         * The aggregation method for calculating totals.
+         * It can be a built-in function (e.g., 'sum') or a custom function.
+         * @default 'sum'
+         */
+        aggregation?:
+            | ChartTooltipTotalsBuiltInAggregation
+            | ((args: ChartTooltipTotalsAggregationArgs) => number | undefined);
```
kuzmadom, line 131: Perhaps the aggregation for the total does not necessarily have to return a number? There may also be text.

src/components/Tooltip/DefaultContent.tsx
```diff
@@ -1,116 +1,93 @@
...
-const getMeasureValue = ({
-    data,
-    xAxis,
-    yAxis,
-    valueFormat,
-}: {
-    data: TooltipDataChunk[];
-    xAxis?: ChartXAxis | null;
-    yAxis?: ChartYAxis;
+function RowTotals(props: {
```
kuzmadom, line 61: Maybe you should move it to a separate file? To make it easier to read the code

src/components/Tooltip/utils.ts
```diff
@@ -0,0 +1,196 @@
...
+    totals?: ChartTooltip['totals'];
+    xAxis?: ChartXAxis | null;
+    yAxis?: ChartYAxis;
+}): ChartTooltipTotalsBuiltInAggregation | (() => number | undefined) {
+    const {hovered, totals, xAxis, yAxis} = args;
+
+    const aggregation = totals?.aggregation;
+
+    if (typeof aggregation === 'string') {
+        return aggregation;
```
kuzmadom, line 188: Perhaps it's worth checking for acceptable aggregation here? If not, then we seem to swallow the error and output nothing.

kuzmadom: Снимок экрана 2025-10-02 в 14 14 07

PR #181, opened 2025-10-04
fix: fix bar-y chart layout

korvin89:


src/hooks/useAxisScales/index.ts
```diff
@@ -67,9 +69,57 @@ const filterCategoriesByVisibleSeries = (args: {
...
+    }
+
+    let seriesWithStackingHasAlreadyCounted = false;
+    const offsetMultiplier = preparedBarYSeries.reduce((acc, s) => {
+        let count = 0;
+
+        if (s.stacking && !seriesWithStackingHasAlreadyCounted) {
+            seriesWithStackingHasAlreadyCounted = true;
+            count = 1;
+        } else if (!s.stacking) {
```
kuzmadom, line 101: Is this enough to check for sure?
The series may have a different stack id

kuzmadom (approved):

PR #205, opened 2025-10-14
feat: improve overflowed tooltip rows view

korvin89:


src/components/Tooltip/DefaultTooltipContent/RowWithAggregation.tsx
```diff
@@ -14,13 +14,23 @@ import type {HoveredValue} from './utils';
...
 
-export function RowTotals(props: {
+// label: React.ReactNode;
+//     value: React.ReactNode;
+//     active?: boolean;
+//     className?: string;
+//     color?: string;
+//     striped?: boolean;
+//     style?: React.CSSProperties;
+
```
kuzmadom: should be deleted I suppose?

src/components/Tooltip/DefaultTooltipContent/index.tsx
```diff
@@ -219,14 +269,33 @@ export const DefaultTooltipContent = ({
...
-            }
+                })}
+                {Boolean(restHoveredValues.length) && (
+                    <RowWithAggregation
+                        aggregation={getPreparedAggregation({
+                            hovered: restHovered,
+                            totals,
+                            xAxis,
+                            yAxis,
+                        })}
```
kuzmadom: It turns out that with `totals.enabed: false`, we still count aggregation for collapsed values? I think this is wrong

kuzmadom (commented):
Снимок экрана 2025-10-15 в 11 06 42
Regarding the following improvements: the value on the right now occupies a portion of the column on the left

PR #217, opened 2025-10-22
fix: fix bar-y data lables outside plot boundaries

korvin89:


src/hooks/useShapes/bar-y/prepare-data.ts
```diff
@@ -18,15 +18,81 @@ import type {BarYShapesArgs, PreparedBarYData} from './types';
...
+    if (y - height < 0) {
+        resultY = 0;
+    }
+
+    if (y > boundsHeight) {
+        resultY = boundsHeight;
+    }
+
+    return {x: resultX, y: resultY};
+}
```
kuzmadom: To a shared utility? It looks useful not only for bar-y series

src/hooks/useShapes/bar-y/prepare-data.ts
```diff
@@ -18,15 +18,81 @@ import type {BarYShapesArgs, PreparedBarYData} from './types';
...
+    if (y < 0) {
+        resultY = 0;
+    }
+
+    if (y + height > boundsHeight) {
+        resultY = boundsHeight - height;
+    }
+
+    return {x: resultX, y: resultY};
+}
```
kuzmadom: Same here

PR #227, opened 2025-10-28
fix: y-axis labels should be displayed considering only the visible series

kuzmadom:


src/components/ChartInner/useChartInnerProps.ts
```diff
@@ -2,6 +2,8 @@ import React from 'react';
 
 import type {Dispatch} from 'd3';
 
+import type {ChartSeries} from 'src/types';
```
korvin89: We don't have aliases in tsconfig, it is better to use relative path

korvin89 (approved):

PR #241, opened 2025-11-05
feat: basic funnel chart

kuzmadom:


src/hooks/useSeries/prepare-funnel.ts
```diff
@@ -0,0 +1,64 @@
...
+                dataLabels: {
+                    enabled: get(series, 'dataLabels.enabled', true),
+                    style: Object.assign({}, DEFAULT_DATALABELS_STYLE, series.dataLabels?.style),
+                    html: get(series, 'dataLabels.html', false),
+                    format: series.dataLabels?.format,
+                    align: series.dataLabels?.align ?? 'center',
+                },
+                visible: true,
+                name: dataItem.name,
+                id: `Series ${i}`,
```
korvin89: Maybe its better to use `getUniqId` as we do in other cases?

PR #257, opened 2025-11-11
refactor: change bar-y border rendering

korvin89:


src/hooks/useShapes/bar-y/index.tsx
```diff
@@ -41,21 +41,59 @@ export const BarYSeriesShapes = (args: Args) => {
...
+                        fillX = d.x;
+                        fillWidth = d.width;
+                        fillBorderRadiusRight = 0;
+                    } else if (d.skipBorderEnd) {
+                        // No right border - fill extends to the right edge
+                        // Position was shifted right by halfBorder, fill extends full width from that point
+                        fillWidth = d.width;
+                        fillBorderRadiusRight = 0;
+                    }
+                }
```
kuzmadom: I think such calculations should be in prepare-data, not here. Or put it as a utility function in a separate file.

src/__snapshots__/bar-y-series.visual.test.tsx-snapshots/Bar-y-series-negative-values-with-border-1-chromium-linux.png
kuzmadom: I think the border should be everywhere - around the entire segment.
For example, here are examples in highcharts and echarts:
Снимок экрана 2025-11-11 в 22 38 45
Снимок экрана 2025-11-11 в 22 38 28

PR #263, opened 2025-11-15
feat: add position to the legend configuration (top and bottom options)

arteria32:


arteria32:
@korvin89 @kuzmadom 
Hi, implement the first step - adding a bottom and a top option for the legend position.

src/types/chart/legend.ts
```diff
@@ -71,6 +71,13 @@ export interface ChartLegend {
      * @default false
      * */
     html?: boolean;
+    /**
+     * The position of the legend box within the chart area.
+     *
+     * @experimental Can be changed in the future.
```
kuzmadom: I don't think the tag is needed here? It looks stable enough to use without it. Additional changes in the future, such as adding left and right positioning, do not affect the stability of the API interface.

PR #283, opened 2025-11-25
fix: basic bar-x split

kuzmadom:


kuzmadom:
> Something goes wrong in case of clicking on `Series 2` legend item
> 
fixed

PR #284, opened 2025-11-25
fix: fix zoom in case of min/max usage

korvin89:


src/hooks/useAxisScales/utils.ts
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * Checks whether a domain is zero (when minimum and maximum values are equal).
+ *
+ * This is necessary for cases where exactly one marker needs to be rendered on an axis.
+ * In such cases, it is not allowed to use axis extremums (min/max)
+ * that differ from those in the domain, as this can lead to incorrect visualization
+ * and scale stretching around a single point.
+ */
+export function isZeroDomain(domain: [number, number]) {
+    return domain[0] === domain[1];
```
kuzmadom: for series with a single point and which are not tied to the zero value of the axis (without accumulations, etc.) - scatter or linear diagrams - it looks incorrect. In their case, the domain will consist of one value, but not zero

PR #288, opened 2025-11-26
fix: x-axis max value padding should affect axis scale domain, not range

kuzmadom:
It's like this now:
Снимок экрана 2025-11-26 в 16 45 21

but it should be something like:
Снимок экрана 2025-11-26 в 16 47 03


src/__snapshots__/bar-y-series.visual.test.tsx-snapshots/Bar-y-series-Basic-with-reverse-order-on-X-axis-1-chromium-linux.png
korvin89: I'm not sure about this difference. Current behavior (in main) looks correct:
<details>
<summary>With default maxPadding (0.01)</summary>

Image

</details>
<details>
<summary>With maxPadding 0.2</summary>

Image

</details>
<details>
<summary>With maxPadding 0</summary>

Image

</details>

With the current changes, it looks like the padding is set to zero, although the default value of 0.01 is used in the test

src/__snapshots__/bar-y-series.visual.test.tsx-snapshots/Bar-y-series-Y-axis-linear-percent-stacked-series-1-chromium-linux.png
korvin89: The same thing about `maxPadding` here

PR #339, opened 2025-12-25
feat(Legend): add property `verticalAlign` to legend config

arteria32:


src/types/chart/legend.ts
```diff
@@ -27,6 +27,14 @@ export interface ChartLegend extends ChartLegendItem {
...
      * */
     align?: 'left' | 'center' | 'right';
+
+    /**
+     * The vertical alignment of the legend box within the chart area.
+     * Only applies when position is 'left' or 'right'.
+     *
+     * @default top
+     * */
+    alignVertical?: 'top' | 'center' | 'bottom';
```
kuzmadom: Everything is great, but maybe it would be worth changing the name to `verticalAlign`? This is more expected compared to `alignVertical` in my opinion

kuzmadom (approved):

PR #343, opened 2025-12-25
feat: allow line series to redefine styles for their representation in the range slider

korvin89:


src/types/chart/line.ts
```diff
@@ -32,30 +32,33 @@ export interface LineSeriesData<T = MeaningfulAny> extends BaseSeriesData<T> {
...
 }
 
-export interface LineSeries<T = MeaningfulAny> extends BaseSeries {
-    type: typeof SERIES_TYPE.Line;
-    data: LineSeriesData<T>[];
-    /** The name of the series (used in legend, tooltip etc) */
-    name: string;
+export interface LineSeriesLineStyle {
     /** The main color of the series (hex, rgba) */
     color?: string;
```
kuzmadom: Do we really need to be able to redefine the values of colors or the shapes of lines? Line width or transparency is ok, it may depend on the scale. But the color - it serves as an indicator for identification, in which cases should it be redefined?

kuzmadom (approved):

PR #503, opened 2026-03-21
docs(formatter): add JSDoc with examples to FormatOptions and FormatNumberOptions

korvin89:


src/types/formatter.ts
```diff
@@ -1,14 +1,102 @@
+/**
+ * Base formatting options shared across all formatters (number, bytes, duration).
```
kuzmadom: bytes, duration ?

src/types/formatter.ts
```diff
@@ -1,14 +1,102 @@
...
+ *
+ * @example
+ * // Fixed precision
+ * formatNumber(1234.5678, { precision: 2 }); // "1,234.57"
```
kuzmadom: `formatNumber` - what is it ? If these are examples for external users, they don't have such a method.

PR #550, opened 2026-04-13
feat(format-number): add declarative `units` option for custom unit scales

korvin89:


docs/diplodoc/pages/guides/tooltip.md
```diff
@@ -160,9 +215,13 @@ series-level setting takes precedence over the chart-level one for that series o
...
         name: 'Bandwidth',
-        data: [/* ... */],
+        data: [/* y in bytes */],
         tooltip: {
-          valueFormat: {type: 'custom', formatter: formatBytes},
+          valueFormat: {
+            type: 'number',
+            precision: 1,
+            units: {base: 1024, labels: ['B', 'KB', 'MB', 'GB', 'TB']},
+          },
```
kuzmadom, line 113: The tooltip documentation contains too much general formatting info. I suggest moving all formatting details (units, custom formats, etc.) to a separate article, and replacing it in the tooltip doc with a brief note like "Formatting works the same way as elsewhere — see [Formatting]." Keep only tooltip-specific things here, like overriding the format for a specific series.

src/libs/format-number/format-number.test.ts
```diff
@@ -5,6 +5,18 @@ import {formatNumber} from '.';
...
+    base: 1024,
+    labels: ['B', 'KB', 'MB', 'GB', 'TB'],
+};
+
+const TIME: FormatNumberOptions['units'] = [
+    {factor: 1, label: 's'},
+    {factor: 60, label: 'min'},
+    {factor: 3600, label: 'h'},
+    {factor: 86400, label: 'd'},
+];
```
kuzmadom: Possible to move this to shared constants? In the future, it could be reused for common dictionaries — when defining formats via units.

src/libs/format-number/types.ts
```diff
@@ -1,14 +1,71 @@
...
+    /** Internal rendering hint for axis label layout. Not intended for public use. */
     labelMode?: string;
+}
+
+/** Single entry of a custom unit scale. */
+export type FormatUnitScaleEntry = {
+    /** Positive multiplier applied to the value. The label wins when `|value| / factor >= 1`. */
+    factor: number;
+    /** Suffix rendered after the scaled value (e.g. `'KB'`). Use an empty string to suppress the suffix and the delimiter. */
+    label: string;
```
kuzmadom: Maybe it's worth naming this the same as in the general formatting settings — postfix? That way, if a unit ever needs a prefix as well, it will extend naturally without any issues.

src/libs/format-number/types.ts
```diff
@@ -1,14 +1,71 @@
...
+     * Custom unit scale. When set, fully overrides `unit`.
+     * Unit selection is always automatic: the entry with the largest `factor` such that `|value| / factor >= 1`,
+     * clamped to the smallest entry for very small values.
+     */
+    units?: FormatUnitScale;
+    /**
+     * String placed between the scaled value and the unit label when `units` is set.
+     * Defaults to a locale-aware space. Use `''` to render without a separator (e.g. `'1.5h'`).
+     */
+    unitDelimiter?: string;
```
kuzmadom: Maybe unitDelimiter should be moved into the units block? It doesn't really make sense without it.

src/libs/format-number/types.ts
```diff
@@ -1,14 +1,71 @@
...
+     * - `'auto'` — picks the most appropriate unit automatically (`k`, `m`, `b`, `t`).
+     * - `'k'` — thousands (÷ 1 000).
+     * - `'m'` — millions (÷ 1 000 000).
+     * - `'b'` — billions (÷ 1 000 000 000).
+     * - `'t'` — trillions (÷ 1 000 000 000 000).
+     * - `null` — no unit suffix.
+     *
+     * @deprecated Use `units` for custom scales. This option is fully ignored when `units` is set.
+     */
     unit?: 'auto' | 'k' | 'm' | 'b' | 't' | null;
```
kuzmadom, line 82: `unit` is currently marked as deprecated, but moving to `units` we essentially lost the ability to explicitly set the desired level — for example, previously you could do something like 0.3K to force everything to display in thousands. Now only auto is left, which doesn't allow setting the level explicitly. 
Maybe it's worth adding a corresponding setting to `units` to support this?

src/libs/format-number/format-number.test.ts
korvin89: Fair point on reuse, but I'd rather keep these as local test fixtures for now. Exposing them as shared constants (e.g. FORMAT_UNIT_PRESETS.bytes) would add public API surface that no real consumer has asked for yet — it's speculative reuse, and we'd be locked into the naming, shape, and language (e.g. 'KB' vs 'КБ', 'min' vs 'м') from day one.

My preference is to wait until an actual callsite needs them: if a user files an issue asking "how do I format bytes?" we can extract a preset at that point, informed by what they actually want. Until then, the declarative units option is already compact enough to write inline, and the test fixtures are short enough that duplicating them locally doesn't hurt.

src/libs/format-number/types.ts
korvin89: Good point — you're right that the explicit-level behavior was lost when unit was deprecated. However, the new units option already supports this use case without a dedicated setting: a single-entry scale acts as a locked level, regardless of the value's magnitude.
```js
// Legacy: force thousands
{unit: 'k', precision: 1}
// → 300 renders as "0.3K"
```
```js
// New equivalent: single-entry scale
{units: {scale: [{factor: 1000, postfix: 'K'}]}, precision: 1}
// → 300 renders as "0.3 K"
```

kuzmadom: Looks good overall — it's great that everything is configurable. One thing to consider though: if a user needs to set the same unit everywhere (or auto), they now have to repeat `{units: {base: 1000, postfixes: ['K' | 'M' | 'B' | 'T']}}` across axes/tooltip/labels instead of just `{unit: 'auto'}`. It got more verbose. Are we okay with that?

korvin89: The fact that `{unit: 'k', precision: 1}` will be defined as `{units: {scale: [{factor: 1000, postfix: 'K'}]}, precision: 1}` seems fine — the notation became more verbose, but on the other hand more flexible. The fact that we need to define this in multiple places when we want the same formatting everywhere — that's not ok. It feels like we need to define one common place (maybe in `series.options` or somewhere else) where we would override the default `valueFormat`. What do you think about this?

kuzmadom (dismissed):

kuzmadom (approved):

PR #579, opened 2026-04-24
refactor(core): isolate src/core from React dependencies

kuzmadom:
## Summary                                                                                                                              
                                                                                                                                          
  - Add `src/core/types/css.ts` — React-free `CSSProperties` type alias over `csstype.Properties` (replaces `React.CSSProperties` in core)
  - Add `src/core/types/renderer.ts` — React-free `RendererElement` and `SVGTextAttributes` interfaces (replaces `React.ReactElement` /  `React.SVGTextElementAttributes` in core)                                        
  - Move `getSeriesClipPathId` from `src/hooks/useShapes/utils.ts` to `src/core/shapes/utils.ts`; the hook re-exports it for backward compat                                 
  - Replace all `React.*` type references inside `src/core/` with the new framework-agnostic equivalents                                  
  - Add `csstype` as an explicit dependency (was already a transitive dep via React)
                                                                                                                                          
  `src/core/` now has zero React imports — prerequisite for the upcoming plugin architecture where plugin code must be framework-agnostic.


src/hooks/useShapes/HtmlLayer.tsx
```diff
@@ -34,13 +34,13 @@ export const HtmlLayer = (props: Props) => {
...
                 const scope = item.scope ?? 'plot';
-                const style: React.CSSProperties = {
+                const style = {
                     ...item.style,
                     color: item.style?.color ?? item.style?.fontColor,
                     position: 'absolute',
                     left: item.x,
                     top: item.y,
-                };
+                } as React.CSSProperties;
```
korvin89: Why do we need this change? Predefined type looks better in this case

package.json
```diff
@@ -71,6 +71,7 @@
     "@gravity-ui/date-utils": "^2.5.4",
     "@gravity-ui/i18n": "^1.6.0",
     "@gravity-ui/icons": "^2.15.0",
+    "csstype": "^3.2.3",
```
korvin89, line 74: Could we put it into dev section like all the types?

kuzmadom: csstype can't be moved to devDependencies. 
The CSSProperties type (which is a direct re-export of Properties from csstype) is used in public API types like HtmlItem.style, which is exported via export * from './types' in index.ts. Since the build uses plain tsc with declaration: true (no type bundling/inlining), the compiled .d.ts files will contain a direct reference to csstype. 
If it were in devDependencies, consumers would get a "Cannot find module 'csstype'" error from TypeScript when resolving the package's types.

korvin89 (approved):

PR #582, opened 2026-04-26
fix(zoom): correct stacked bar-x/bar-y/area rendering under xy zoom

korvin89:
## Problem                                                                                                                        
                                                                                                                                    
  Stacked series under XY zoom rendered incorrectly — segments were lost or anchored to wrong stack positions. Two independent bugs:
                                                                                                                                    
  **1. Per-point value-axis filter dropped stack segments.** `getZoomedSeriesData` checks each point's individual value against the zoom range. For stacked series the chart-space position of a segment is the *cumulative* stack sum, not the raw `point.x` / `point.y`, so segments with small individual values fell outside the zoom range and got dropped — leaving the remaining ones to re-anchor from zero.                                                                                                              
                                                                                                                                    
  **2. `prepareBarYData` direction check used pixel coords as data sign.** The "positive or negative side of base?" branch was `xValue > xLinearScale.range()[0]`. Without zoom `range[0] = 0`, so it accidentally worked as `xValue > 0`. With zoom, `getXScaleRange` adds 2% padding and `range[0]` becomes `0.02·boundsWidth`. Any positive value smaller than that padding got misclassified into `negativeStack` and rendered far off-canvas left.                                                              
   
  ## Fix                                                                                                                            
                  
  `src/core/zoom/zoom.ts` — skip the value-axis filter for stacked series:
                                          
  - `bar-x`, `area` → value axis is Y, skip Y filter
  - `bar-y` → value axis is X, skip X filter

### DIFF
--- a/src/core/shapes/bar-y/prepare-data.ts
+++ b/src/core/shapes/bar-y/prepare-data.ts
@@ -137,10 +137,6 @@ export async function prepareBarYData(args: {
                 const borderWidth = barSize > s.borderWidth * 2 ? s.borderWidth : 0;
                 const isFirstInStack = xValueIndex === 0;
                 const isLastStackItem = xValueIndex === sortedData.length - 1;
-                // Whether the bar extends toward higher pixel coords from base.
-                // Depends on both data sign and axis order (e.g. reverse). Doing
-                // the comparison in pixel space keeps it correct under zoom too,
-                // where range[0] is no longer 0 due to zoom padding.
                 const extendsRight = xLinearScale(xValue) > baseValue;
                 // Calculate position with border compensation
                 // Border extends halfBorder outward from the shape, so we need to adjust position
--- a/src/core/shapes/bar-y/prepare-data.ts
+++ b/src/core/shapes/bar-y/prepare-data.ts
@@ -42,7 +42,6 @@ export async function prepareBarYData(args: {
     const stackGap = seriesOptions['bar-y'].stackGap;
     const xLinearScale = xScale as ScaleLinear<number, number>;
     const yLinearScale = yScale as ScaleLinear<number, number> | undefined;
-    const [baseRangeValue] = xLinearScale.range();
 
     if (!yLinearScale) {
         return {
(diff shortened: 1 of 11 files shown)


src/core/zoom/zoom.ts
```diff
@@ -98,6 +98,19 @@ export function getZoomedSeriesData(args: {
...
+        // raw point value would drop segments that still need to participate
+        // in the stack, so skip the value-axis filter and let the axis range
+        // / clip handle visibility:
+        //   - bar-x, area  → value axis is Y, skip Y filter
+        //   - bar-y        → value axis is X, skip X filter
+        const isStacked = 'stacking' in seriesItem && Boolean(seriesItem.stacking);
+        const skipYFilter =
+            isStacked &&
+            (seriesItem.type === SERIES_TYPE.BarX || seriesItem.type === SERIES_TYPE.Area);
+        const skipXFilter = isStacked && seriesItem.type === SERIES_TYPE.BarY;
```
kuzmadom: Can we embed the accumulation direction sign directly in the series (not in the top-level API, but perhaps in the prepared form)?

Right now there are already too many places in the code that branch on the series type — I'd rather not add more. Ideally, I'd like to move toward eliminating those type checks entirely.

src/core/shapes/bar-y/prepare-data.ts
```diff
@@ -138,19 +137,26 @@ export async function prepareBarYData(args: {
                 const borderWidth = barSize > s.borderWidth * 2 ? s.borderWidth : 0;
                 const isFirstInStack = xValueIndex === 0;
                 const isLastStackItem = xValueIndex === sortedData.length - 1;
+                // Whether the bar extends toward higher pixel coords from base.
+                // Depends on both data sign and axis order (e.g. reverse). Doing
+                // the comparison in pixel space keeps it correct under zoom too,
+                // where range[0] is no longer 0 due to zoom padding.
```
kuzmadom: Are you sure you need to leave comments about zoom padding? It looks like something to debug/fix a specific error

kuzmadom (approved):

PR #605, opened 2026-05-07
fix: add entry points to sideEffects to prevent tree-shaking of plugin registration

kuzmadom:
Rollup in consumers' production builds was dropping import './plugins' from dist/esm/index.js because the main entry point was not marked as a side-effect module. Adding index.js (esm and cjs) to sideEffects ensures plugin registration is not eliminated during tree-shaking.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,10 +22,8 @@
   "sideEffects": [
     "*.css",
     "*.scss",
-    "./dist/esm/index.js",
-    "./dist/cjs/index.js",
-    "./dist/esm/plugins/index.js",
-    "./dist/cjs/plugins/index.js"
+    "**/plugins/index.*",
+    "./dist/*/index.js"
   ],
   "files": [
     "dist",
--- a/package.json
+++ b/package.json
@@ -22,7 +22,6 @@
   "sideEffects": [
     "*.css",
     "*.scss",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -22,9 +22,8 @@
   "sideEffects": [
     "*.css",
     "*.scss",
-    "./src/plugins/index.ts",
-    "./dist/esm/plugins/index.js",
-    "./dist/cjs/plugins/index.js"
+    "**/plugins/index.*",
+    "./dist/*/index.js"
```
kuzmadom, line 26: `**/plugins/index.*` — targets only files named `index.*` inside a directory called plugins. This covers the registration barrel in both source (src/plugins/index.ts) and compiled output (`dist/esm/plugins/index.js`, `dist/cjs/plugins/index.js`) without listing each path explicitly.                                                                                                                               
                                                                                                                                            
  `./dist/*/index.js` — the single * matches exactly one path segment, so it covers only `dist/esm/index.js` and `dist/cjs/index.js`. The main entry points must be marked as side-effectful because they contain import './plugins', which bundlers would otherwise drop when tree-shaking named exports.

korvin89 (approved):

PR #610, opened 2026-05-10
fix: don't show null-source points as 0 in tooltip and labels

korvin89:
## Summary

`nullMode: 'zero'` rewrites `null → 0` at `prepareSeriesData` time so the y-axis domain and stack/group baselines stay consistent. The substituted point then flowed through to interactive UI as a real `0` — even when the resulting shape was visually invisible (line/area dropping to baseline, 0-height bar, 0-width bar). Users hovered over empty space and saw a "ghost" tooltip row / data label `"0"` for a value they hadn't authored.

This PR hides the substituted points from tooltip rows, data labels, markers and hover markers in the four series where the 0-fill is **structurally invisible**: `area`, `line`, `bar-x`, `bar-y`. The fix preserves the existing visual layout (stack height, slot reservation, axis range) — it only stops surfacing the placeholder in interactive UI.

`scatter`, `heatmap`, `pie` and `waterfall` are intentionally **not** changed: their `'zero'` substitution renders a visible artefact (marker at `(x, 0)`, colored cell, legend entry, plateau in the connector), and the tooltip/label `"0"` correctly reflects what the user is looking at.

## Implementation

Each fixed series follows the same pattern (codified in [`AGENTS.md`](AGENTS.md)):

1. `prepareSeriesData` keeps the `null → 0` substitution so the axis domain still includes 0.
2. `Prepared<Series>` carries the unmodified input as `originalData` (and `nullMode`).
3. Shape preparation tags points whose `originalData[i]` had a null source value with `excluded?: boolean`.
4. The plugin filters `excluded` at its call sites of shared utilities (`preparePointDataLabels`, `buildHoverMarkerGetter`) and inside its own `getTooltipData` and marker-reduce. **Shared utilities stay ignorant** of `nullMode`/`excluded`.
5. Path generators are untouched — they keep the `y === null` check, so `'skip'` mode still produces gaps.

## Series classification

| Series | `'zero'` 0-fill is | Decision |
|---|---|---|
| area | invisible (path drops to baseline) | fix |
| line | invisible (line drops to 0) | fix |
| bar-x | invisible (0-height bar) | fix |
| bar-y | invisible (0-width bar) | fix |
| scatter | visible (marker at `(x, 0)`) | leave |
| heatmap | visible (cell colored at value 0) | leave |
| pie | visible (legend entry, 0-angle slice) | leave |
| waterfall | visible (plateau in connector) | leave |


AGENTS.md
```diff
@@ -5,3 +5,33 @@
 # Plugin architecture
 
 - Series-specific behavior must live in the plugin, not in shared hooks or utilities. Never add a hardcoded `Set`, `switch`, or constant keyed on series type string in shared code — add a field to `SeriesPlugin` and let each plugin declare its own behavior.
+
+# Null-mode handling
```
kuzmadom, line 9: Quite specific information - does it really have to live in general AGENTS.md ? It seems that this will just fill in the context - it is already embedded in the code and covered with tests. If you want to keep this logic described, then maybe you should create a separate section like technotes?

src/core/shapes/bar-x/prepare-data.ts
```diff
@@ -137,6 +137,21 @@ export const prepareBarXData = async (args: {
...
+    // `nullMode: 'zero'`. The bar still occupies its slot, but tooltip rows
+    // and data labels must skip it.
+    const excludedData = new Set<BarXSeriesData>();
+    series.forEach((s) => {
+        if (s.nullMode !== 'zero') {
+            return;
+        }
+        s.data.forEach((d, i) => {
+            if (s.originalData[i]?.y === null) {
+                excludedData.add(d);
```
kuzmadom, line 150: Perhaps instead of saving originalData in a series, it's worth putting it at each point? Then there won't be an extra few cycles

src/core/shapes/bar-x/prepare-data.ts
```diff
@@ -137,6 +137,21 @@ export const prepareBarXData = async (args: {
     const domain = new Set<string | number>();
     let maxGroupSize = 1;
 
+    // Mark data points whose source y was null and were rewritten to 0 by
+    // `nullMode: 'zero'`. The bar still occupies its slot, but tooltip rows
+    // and data labels must skip it.
+    const excludedData = new Set<BarXSeriesData>();
+    series.forEach((s) => {
+        if (s.nullMode !== 'zero') {
```
kuzmadom, line 145: Why is there only a `zero` check? Are there any other cases where we need to display null in a tooltip?

src/__tests__/area-series.visual.test.tsx
```diff
@@ -120,6 +120,54 @@ test.describe('Area series', () => {
...
+            const svgBox = await getLocatorBoundingBox(svg);
+            await page.mouse.move(
+                Math.round(tickBox.x + tickBox.width / 2),
+                Math.round(svgBox.y + svgBox.height / 2),
+            );
+
+            const tooltip = page.locator('.gcharts-tooltip');
+            await expect(tooltip).toBeVisible();
+            await expect(tooltip.getByText('Series B')).toBeVisible();
+            await expect(tooltip.getByText('Series A')).toHaveCount(0);
```
kuzmadom, line 169: Asserting the absence of an element by its text is highly prone to false positives. If the copy changes in the future, this test will silently pass even if the element incorrectly appears on the screen.
Since this element is conditionally hidden, maybe we should rewrite this as a visual/screenshot test? It would reliably verify that the component looks correct without this text. 
Alternatively, we could assert the positive state (what should be visible instead) with all tooltip rows.But in this case, it is worth adding a comment explaining why we expect such a composition and what exactly should not be.

korvin89:
Will be fixed here https://github.com/gravity-ui/charts/pull/612

PR #630, opened 2026-06-04
ci: cross-version Playwright runs on release-please PRs

korvin89:
## Summary

Default CI keeps running on React 18. On release-please PRs additionally run Playwright (visual + perf) against the edges of our peer range (React 17 and React 19), so a release can't go out if cross-version regressions slipped in.

- `scripts/install-react.mjs` — swaps `react`, `react-dom`, `@types/react`, `@types/react-dom`, and a compatible `@testing-library/react` for React 17 or 19. Uses `--no-save` and backs up `package-lock.json` so the lockfile isn't disturbed (which keeps `@playwright/test` pinned to the version the Docker image expects).
- `scripts/playwright-docker.sh` — adds opt-in `REACT_VERSION=17|19`: isolated `node_modules` cache per version (`...-react17|19`), `clear-cache` wipes all variants, `REACT_VERSION` is propagated into the container so the Vite alias in `playwright.config.ts` activates.
- React 17 doesn't ship `react-dom/client`, but Playwright CT's `registerSource` imports `createRoot` from there. Stay on modern Playwright 1.56.1-jammy for all versions and alias `react-dom/client` to a thin compat shim (`playwright/shims/react-dom-client.ts`) when `REACT_VERSION=17`. The shim implements `createRoot`/`hydrateRoot` on top of `ReactDOM.render`.
- New npm scripts: `playwright:react17`, `playwright:react19`, `playwright:docker:react17`, `playwright:docker:react19`. Existing aliases (`playwright`, `playwright:docker`, `playwright:docker:update`, `playwright:docker:clear-cache`) untouched and continue to run on the default React 18 dev set.
- `pr-visual-tests.yml`: default `visual_tests` + `visual_tests_perf` unchanged. Adds gated `visual_tests_react_17|19` and `visual_tests_perf_react_17|19` with `if: startsWith(github.head_ref, 'release-please')`.
- `pr-visual-tests-report.yml`: backwards-compatible — uploads the default report to the old S3 paths (`pulls/<PR>/` and `pulls/<PR>/perf/`); conditionally also uploads `react17` / `react19` subpaths and includes them in the PR comment when present.

## Notes

- Master React stays at 18 in this PR. **Bump to React 19 in devDeps is deferred to a separate PR** — running it locally surfaces ~240 typecheck errors in `__stories__/` (`AnnotatedStoryFn<ReactRenderer, ...>` is no longer JSX-valid under @types/react@19), and that cleanup belongs in its own change.
- After a local run of `playwright:react17|19`, run `npm ci` to restore the dev set.
- **Report comment is broken on this PR by design.** GitHub runs `workflow_run`-triggered workflows from the **default branch's** copy of the file. The updated `pr-visual-tests-report.yml` only takes effect after this PR (or a separate report-only PR) is merged to `main` — until then the report job runs main's old script and fails because the artifact paths shift. The new report code itself has been reviewed locally; it'll fire on the next PR after this one merges.


package.json
```diff
@@ -63,9 +63,15 @@
     "playwright:perf": "playwright test --project=chromium-perf --workers=1 --no-deps --config=playwright/playwright.config.ts",
     "playwright:update": "npm run playwright -- -u",
     "playwright:clear-cache": "rm -rf ./playwright/.cache",
+    "playwright:react17": "node scripts/install-react.mjs 17 && REACT_VERSION=17 npm run playwright",
+    "playwright:react18": "node scripts/install-react.mjs 18 && npm run playwright",
```
kuzmadom: Почему здесь нет REACT_VERSION?

package.json
```diff
@@ -63,9 +63,13 @@
     "playwright:perf": "playwright test --project=chromium-perf --workers=1 --no-deps --config=playwright/playwright.config.ts",
     "playwright:update": "npm run playwright -- -u",
     "playwright:clear-cache": "rm -rf ./playwright/.cache",
+    "playwright:react17": "node scripts/install-react.mjs 17 && REACT_VERSION=17 npm run playwright",
+    "playwright:react19": "node scripts/install-react.mjs 19 && npm run playwright",
```
kuzmadom: there is no env variable REACT_VERSION - is it ok?

scripts/install-react.mjs
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/env node
+import {spawnSync} from 'node:child_process';
+import {readFileSync, writeFileSync} from 'node:fs';
+import {resolve} from 'node:path';
+import {argv, cwd, exit, stderr, stdout} from 'node:process';
+
+// React 18 is the default in package.json; this script swaps the dev
```
kuzmadom, line 7: It might be worth writing a short (just really short - one paragraph somewhere in tech notes) instruction for the next react updates. Now it turns out that there is a lot of knowledge about the default version scattered around the code/comments

kuzmadom (dismissed):

package.json
korvin89: Yes, it's intentional. REACT_VERSION currently affects only one thing — the react-dom/client → shim alias in playwright/playwright.config.ts, and that alias activates only when REACT_VERSION=17 (React 17 doesn't ship react-dom/client). For React 19 the shim isn't needed, so the env var would  be a no-op.

kuzmadom (approved):

PR #641, opened 2026-06-19
refactor(validation): move per-type validation into series plugins

korvin89:
> 🤖 AI generated

## Description
`validateData` had a switch over the series type, so adding a new type meant
editing a central file. Type-specific validation now lives in the type's own
plugin, keeping the change local. Part of the ongoing pluginization effort.

## What changed
- `SeriesPlugin`: optional `validate?(args)` method + `ValidateSeriesArgs` type.
  Core loops over series and delegates via `getSeriesPlugin(type).validate?.(...)`.
- The registry is now the source of truth for available types: `hasSeriesPlugin` /
  `getRegisteredSeriesTypes` replace the hardcoded `SERIES_TYPE` list in validation.
- Shared XY helpers extracted into `core/validation/helpers.ts`.
- `validate()` added to area, bar-x, bar-y, line, scatter, pie, treemap.
  Single-consumer logic (pie value check, treemap structure check) lives in its
  own plugin rather than in shared helpers.
- Removed `validateSeries` / `countSeriesByType` / `validateStackingAreaNullMode`.
- The validation test registers plugins (`import '../../../plugins'`) so the
  registry is populated.

## Behavior
Unchanged — same error codes and messages. The only difference is that the type
list in the `invalid-series-type` message now follows plugin registration order.

### DIFF
--- a/src/core/validation/helpers.ts
+++ b/src/core/validation/helpers.ts
@@ -3,20 +3,18 @@ import get from 'lodash/get';
 import {CHART_ERROR_CODE, ChartError} from '../../libs';
 import {DEFAULT_AXIS_TYPE} from '../constants';
 import {i18n} from '../i18n';
-import type {
-    AreaSeries,
-    BarXSeries,
-    BarYSeries,
-    ChartXAxis,
-    ChartYAxis,
-    LineSeries,
-    ScatterSeries,
-} from '../types';
+import type {ChartXAxis, ChartYAxis} from '../types';
 
-export type XYSeries = ScatterSeries | BarXSeries | BarYSeries | LineSeries | AreaSeries;
+type XYDataPoint = {x?: unknown; y?: unknown};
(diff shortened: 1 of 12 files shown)


src/core/validation/helpers.ts
```diff
@@ -0,0 +1,298 @@
...
+    AreaSeries,
+    BarXSeries,
+    BarYSeries,
+    ChartXAxis,
+    ChartYAxis,
+    LineSeries,
+    ScatterSeries,
+} from '../types';
+
+export type XYSeries = ScatterSeries | BarXSeries | BarYSeries | LineSeries | AreaSeries;
```
kuzmadom: The mecanism of plugins is needed precisely in order not to list specific types of series in such a case. There can be no specific series here and below

kuzmadom (approved):

PR #642, opened 2026-06-23
refactor(color): compute continuous color domain via series plugins

korvin89:
> 🤖 AI generated

## Description
`getDomainForContinuousColorScale` switched over the series type, so a new type
meant editing `core/utils/color.ts`. The per-point color value now comes from the
type's own plugin. Part of the ongoing pluginization effort.

## What changed
- `SeriesPlugin`: optional `getColorValue?(data)`. Core delegates via
  `getSeriesPlugin(type).getColorValue?.(...)` instead of switching.
- Implemented in the 10 color-capable plugins, same field as before: `y`
  (line/area/bar-x/waterfall/scatter), `x` (bar-y), `value` (pie/heatmap/funnel),
  `|x1 - x0|` (x-range). treemap/sankey/radar keep no method and still throw.
- Added a unit test (registers plugins via `import '../../../plugins'`).

## Behavior
Unchanged — same mapping, same `[min, max]`, same error for unsupported types.

### DIFF
--- a/src/core/series/plugin.ts
+++ b/src/core/series/plugin.ts
@@ -67,6 +67,8 @@ export interface ValidateSeriesArgs<T = ChartSeries> {
 }
 
 export interface SeriesPlugin<T extends ChartSeries = ChartSeries> {
+    // --- Metadata ---
+
     /** Unique series type identifier (e.g. `'line'`, `'bar-x'`). Used for plugin lookup and CSS class generation. */
     type: T['type'];
     /**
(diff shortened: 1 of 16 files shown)


src/core/series/plugin.ts
```diff
@@ -81,6 +81,12 @@ export interface SeriesPlugin<T extends ChartSeries = ChartSeries> {
      * Should throw a `ChartError` on invalid input. Omit for types that need no validation.
      */
     validate?(args: ValidateSeriesArgs<T>): void;
+    /**
+     * Returns the numeric value of a data point used to build the domain of a continuous color scale.
+     * Called by `getDomainForContinuousColorScale` over the raw series data (before shape data is prepared).
+     * Omit for types that do not support a continuous color scale (e.g. treemap, sankey, radar).
+     */
+    getColorValue?(data: T['data'][number]): number;
```
kuzmadom: It might be beneficial to group these methods into logical blocks, similar to how it's done with the tooltip. Since the list of functions is growing, adding a bit of structure would prevent the flat list from becoming difficult to read.

src/core/utils/__tests__/color.test.ts
```diff
@@ -0,0 +1,129 @@
+// Populate the series registry so getSeriesPlugin works (see plugin registration gotcha).
```
kuzmadom: Do we have to write this for all tests? Can you put it in a general setup?

src/core/utils/__tests__/color.test.ts
```diff
@@ -0,0 +1,129 @@
...
+import '../../../plugins';
+import type {ChartData} from '../../../types';
+import {getDomainForContinuousColorScale} from '../color';
+
+type Series = ChartData['series']['data'];
+
+describe('utils/color/getDomainForContinuousColorScale', () => {
```
kuzmadom: It might be better to avoid tying test names to file paths. When directories change, keeping test names in sync can be easily overlooked, which might lead to outdated or confusing test names in the future.

src/plugins/bar-y/index.ts
```diff
@@ -49,6 +49,7 @@ export const barYPlugin: SeriesPlugin<BarYSeries> = {
         validateXYSeries({series, xAxis, yAxis});
         validateStacking({series});
     },
+    getColorValue: (d) => Number(d.x),
```
kuzmadom: If colorValue is strictly meant to be a number, it might be better practice to handle the type casting outside the plugin closure rather than inside it.

src/core/series/plugin.ts
```diff
@@ -67,26 +67,47 @@ export interface ValidateSeriesArgs<T = ChartSeries> {
...
     prepareShapeData(
         args: PrepareShapeDataArgs,
     ): PrepareShapeDataResult | Promise<PrepareShapeDataResult>;
+
+    // --- Rendering ---
+
     /** Renders shapes into the provided SVG `<g>` element using D3. May return a cleanup function. */
     renderShapes(args: RenderShapesArgs): (() => void) | void;
+
+    // --- Tooltip ---
```
kuzmadom, line 109: Why are there these comments? In the api, it looks more like something unhealthy. Everything should be clear/well-grouped by object keys

kuzmadom (approved):

PR #650, opened 2026-07-27
fix: dataLabels stop for the rest of the circle after connector bend …

### LINKED ISSUE #649: Pie: dataLabels stop for the rest of the circle after connector bend >90° (shouldStopLabelPlacement)
## Description

On a pie chart with `dataLabels.enabled`, when the chart is small / labels are dense (especially multiline HTML labels), labels are shown only for roughly **half of the circle**. The other half has no labels. Enlarging the chart makes all labels appear.

This comes from `shouldStopLabelPlacement` in `src/core/shapes/pie/prepare-data.ts`: when adjusting a label angle to resolve overlap, if the connector inscribed angle exceeds ~90°, placement stops for **all remaining** segments — not just the conflicting one.

## Expected

- Prefer skipping / hiding only the conflicting label(s), or keep placing later labels that still fit.
- Or at least honor `dataLabels.allowOverlap: true` so that `shouldStopLabelPlacement` does not drop the rest of the labels.

## Actual

Once `shouldStopLabelPlacement` is set to `true`, every subsequent segment is left without a label, which visually looks like “exactly one half of the pie has labels”.

Relevant logic (simplified):

```ts
if (inscribedAngle > 90) {
    shouldAdjustAngle = false;
    shouldStopLabelPlacement = true;
}
// ...
if (!isLabelOverlapped && label.maxWidth > 0 && !shouldStopLabelPlacement) {
    labels.push(label);
}
```

Note: `allowOverlap: true` only affects `isLabelOverlapped`; it does **not** bypass `shouldStopLabelPlacement`.

## Steps to reproduce

1. Render a pie with many segments and HTML multiline labels, e.g.:

```ts
{
  series: {
    data: [{
      type: 'pie',
      dataLabels: { enabled: true, html: true },
      data: Array.from({ length: 12 }, (_, i) => ({
        name: `Segment ${i + 1}`,
        label: `Segment ${i + 1}<br>8.33%`,
        value: 1,
      })),
    }],
  },
}
```

2. Mount in a small container (e.g. ~200×200).
3. Observe labels only on part of the circle.
4. Enlarge the container — all labels appear.

## Environment

- `@gravity-ui/charts` (reproduced against current `main` label layout in `prepare-data.ts`)
- Used via `@gravity-ui/chartkit` gravity-charts plugin

## Suggested direction

- Do not globally abort label placement for the rest of the pie when one connector would bend too much; skip that label (or that adjustment) only.
- When `allowOverlap` is `true`, ignore `shouldStopLabelPlacement` (or never set it).

catakot:
fix https://github.com/gravity-ui/charts/issues/649


src/core/shapes/pie/prepare-data.ts
```diff
@@ -372,7 +371,6 @@ export function preparePieData(args: Args): Promise<PreparedPieData[]> {
 
                             if (inscribedAngle > 90) {
```
kuzmadom, line 372: Previously, setting `shouldStopLabelPlacement = true` also prevented the current label from being added because of the final `!shouldStopLabelPlacement` check. This accidentally acted as a safeguard against using an invalid placement.

  With this PR, the global stop is correctly removed, but the label has already been mutated to `newAngle`, `newX`, and `newY` before `inscribedAngle` is checked.

  If the overlap is resolved on the same iteration where the angle exceeds `90°`, the following overlap check sets `overlap = false`, and the label is added using the position that was just considered invalid. This may produce excessively bent connectors or connectors intersecting the pie.

  Could we keep a local placement-validity flag, or calculate candidate coordinates first and commit them only when the connector angle is valid?

  If no valid position is found, only the current label should be skipped, while placement should continue for the remaining segments.

kuzmadom (commented):
Could we add regression tests for #649?

  Please cover a small pie with dense multiline HTML labels and verify that a failed placement skips only the conflicting label, while subsequent labels are still rendered.

  It would also be useful to verify that `allowOverlap: true` renders all labels.

  The four existing **“Connectors should not intersect with circle”** visual tests currently fail, so their diffs should be reviewed before updating the snapshots.

kuzmadom (approved):

PR #651, opened 2026-07-28
fix(pie): prevent segments from collapsing at specific chart sizes

kuzmadom:
Pie segments with very small fractional values could collapse because D3 rounded SVG arc coordinates to three decimal places.
The path precision is increased to six decimal places for segments and halos. This is sufficient for visible chart radii; the issue could only recur below approximately 0.5px, where the chart is not visually renderable.


src/core/shapes/pie/renderer.ts
```diff
@@ -15,6 +15,21 @@ import {setEllipsisForOverflowTexts} from '../../utils';
 import type {PieLabelData, PreparedPieData, SegmentData} from './types';
 
 const b = block('pie');
+const ARC_PATH_DIGITS = 6;
+
+interface ArcGeneratorWithDigits {
+    digits(digits: number): unknown;
+}
+
+export function getPieArcPath(d: PieArcDatum<SegmentData>, outerRadius: number) {
```
korvin89: 🤖 **AI generated**

**should-fix** — `getPieArcPath` is exported, but I couldn't find any consumer outside this module — just the two call sites below (lines 75 and 87), and `src/index.ts` doesn't re-export it either.

If the export is there to make the function testable, it would be great to cash that in with a unit test — the repo already has jest and the `__tests__` convention (e.g. `src/core/utils/__tests__/format.test.ts`). Something like this pins the regression directly:

```ts
it('keeps arc endpoints distinct for a near-full-circle segment', () => {
    const path = getPieArcPath(segmentWithTinySibling, 250);
    // with digits(3) the arc endpoint collapses onto the start point
    // and the browser drops the `A` command entirely
    expect(path).not.toMatch(/^M(\S+?)A[^A]*,\1(?:L|Z)/);
});
```

That's a much more precise guard than the screenshot, which only catches the bug as a side effect and is sensitive to any layout change. If a unit test isn't planned, then maybe just drop the `export`?

src/__tests__/pie-series.visual.test.tsx
```diff
@@ -33,6 +33,30 @@ test.describe('Pie series', () => {
...
 
+    test('Should render segments with very small fractional values', async ({mount}) => {
+        const data: ChartData = {
+            series: {
+                data: [
+                    {
+                        type: 'pie',
+                        dataLabels: {enabled: true},
+                        data: [
+                            {name: 'Main', value: 100},
```
korvin89, line 44: 🤖 **AI generated**

**nit** — `0.000025` is doing a lot of load-bearing work here, and it's not obvious why. I measured the window in which this test actually reproduces the bug, and it's narrow: at a rendered radius of ~250px only values in roughly `[1.6e-5, 3.1e-5]` trigger the collapse, and at ~286px it shrinks to `[1.6e-5, 2.7e-5]`.

Below the lower edge, d3-path takes its full-circle branch (its own `epsilon = 1e-6`) and renders correctly even with the bug present. Above the upper edge, 3-digit rounding no longer merges the endpoints. So e.g. `0.0000025` or `0.00025` would both make this test pass on the buggy code.

Would you mind adding a short comment on the value? Something like:

```ts
// The angle of this segment must sit just above d3-path's epsilon (1e-6 rad),
// so d3 doesn’t fall back to its full-circle branch, and below 0.0005 / radius,
// so 3-digit rounding still collapses the arc endpoints. Changing this value
// (or the chart size below) can silently stop covering the regression.
{name: 'Small', value: 0.000025},
```

src/core/shapes/pie/renderer.ts
```diff
@@ -15,6 +15,21 @@ import {setEllipsisForOverflowTexts} from '../../utils';
 import type {PieLabelData, PreparedPieData, SegmentData} from './types';
 
 const b = block('pie');
+const ARC_PATH_DIGITS = 6;
```
korvin89, line 20: 🤖 **AI generated**

**nit** — `6` is the result of some genuinely non-obvious math, and right now the reasoning only lives in the PR description, where it'll get lost. Could we bring a short version of it into the code? For example:

```ts
// d3 rounds SVG path coordinates to 3 decimals by default, which can collapse
// the endpoints of a near-full-circle arc into a single point — the browser then
// drops the `A` command and the segment disappears. 6 decimals push that failure
// below a ~0.5px radius, where the chart isn't renderable anyway.
const ARC_PATH_DIGITS = 6;
```

This file's neighbours already document this kind of thing — see the doc comments on `getInscribedAngle` in `utils.ts` and `getLabelPosition` in `prepare-data.ts`.

korvin89 (commented):
🤖 **AI generated**

Left a few inline comments below. The labels I use:

- **blocker** — must be fixed before merge (correctness / security / data loss)
- **should-fix** — a real problem; worth fixing before or right after merge
- **nit** — minor, entirely the author’s discretion
- **question** — I need more information to judge; not an assertion of a defect

korvin89 (approved):

PR #659, opened 2026-08-10
fix(pie): labels overlaps with first label

catakot:


kuzmadom (commented):
~20 snapshots updated, including cases that don't look HTML‑label related (`Basic`, `Min-radius-*`, `Transform-scale`, `nullMode-*`, `With-limited-dimensions-*`). Are all of these updates actually necessary? If the fix only targets the wrap‑around collision with the first label, most of these shouldn't shift. Please double‑check whether the broader overlap‑geometry change is intended, and if not, narrow the fix so unrelated snapshots stay untouched.

src/core/shapes/pie/prepare-data.ts
```diff
@@ -372,7 +398,10 @@ export function preparePieData(args: Args): Promise<PreparedPieData[]> {
                             if (inscribedAngle > 90) {
                                 // Skip only this label; do not commit an invalid connector position.
                                 shouldAdjustAngle = false;
-                            } else if (!isLabelsOverlapping(prevLabel, label, dataLabels.padding)) {
+                            } else if (firstLabel && areLabelsOverlapping(firstLabel, label)) {
```
kuzmadom: Looks redundant with the final `if (!overlap && firstLabel)` check at line 413. The `newAngle % FULL_CIRCLE > labels[0].angle` guard already prevents wrapping past `firstLabel`, so if branch 3 exits with `overlap = false`, the final check immediately flips it back to `true`. The label is discarded either way, intermediate mutations don't leak. Saves 1–2 iterations at best, but adds a third branch to a dense loop and the comment is misleading. Suggest removing it.

src/core/shapes/pie/prepare-data.ts
```diff
@@ -218,6 +218,24 @@ export function preparePieData(args: Args): Promise<PreparedPieData[]> {
...
 
         const shouldUseHtml = dataLabels.html;
+
+        /**
+         * `isLabelsOverlapping` expects the geometry of an SVG text: `x` resolved through
+         * `textAnchor` and `y` pointing at the text baseline (the bottom of the box).
+         * HTML labels store the top-left corner of the element instead, so their bounding box
+         * has to be translated into that form before any comparison.
+         */
+        const getOverlapBox = (label: PieLabelData): PieLabelData => {
```
korvin89: This looks like an unnecessary abstraction - better to use the existing isLabelsOverlapping helper and make the changes there.

src/core/shapes/pie/prepare-data.ts
```diff
@@ -243,6 +261,14 @@ export function preparePieData(args: Args): Promise<PreparedPieData[]> {
...
             }
             const prevLabel = labels[labels.length - 1];
+            /**
+             * Labels are placed clockwise, so the pie closes the circle: the last labels are
+             * neighbours of the very first one and can collide with it while `prevLabel` stays
+             * far away. Skipped when the first label is also the previous one — in that case
+             * the regular `prevLabel` check already covers it, and moving the label forward
+             * increases the distance between them instead of reducing it.
+             */
+            const firstLabel = labels.length > 1 ? labels[0] : undefined;
```
korvin89, line 270: This whole comment seems unnecessary - the code speaks for itself.

src/core/shapes/pie/prepare-data.ts
catakot: There was a problem with the types, which forced me to make more changes. Could you check please?

src/core/shapes/pie/prepare-data.ts
catakot: done

catakot:
> ~20 snapshots updated, including cases that don't look HTML‑label related (`Basic`, `Min-radius-*`, `Transform-scale`, `nullMode-*`, `With-limited-dimensions-*`). Are all of these updates actually necessary? If the fix only targets the wrap‑around collision with the first label, most of these shouldn't shift. Please double‑check whether the broader overlap‑geometry change is intended, and if not, narrow the fix so unrelated snapshots stay untouched.

korvin89 (dismissed):

kuzmadom (dismissed):

korvin89 (approved):

PR #660, opened 2026-08-11
feat: auto placement of dataLabels for line series

### LINKED ISSUE #657: Add automatic data label placement for line series
## Problem

  Line-series data labels are currently placed above their points. Their position cannot be configured, so labels may overlap neighboring lines when multiple series are displayed.

  ## Proposal

  Add an opt-in placement option to line-series data labels:

```
  dataLabels: {
      placement: {
          type: 'auto',
          order: ['top', 'bottom', 'right', 'left'],
          fallback: 'hide',
      },
  }
```

  With type: 'auto', the chart should:

  1. Try candidate positions in the configured order.
  2. Select the first position that does not overlap any visible line.
  3. Apply fallback if no suitable position is found.

  The initial implementation may support only automatic placement with a built-in order and hide fallback. The object shape leaves room for configurable ordering and fallback behavior later.

  This behavior must be opt-in. When placement is omitted, the current label positioning should remain unchanged.

  ## API location

  This feature is specific to line series and should be declared under LineSeries.dataLabels, not BaseDataLabels. The positioning logic should remain inside the line plugin.

## Related approaches

  - [Vega's label transform](https://vega.github.io/vega/docs/transforms/label/) can try multiple anchor positions and avoid intersections with specified line marks.
  - [MapLibre's variable label placement](https://maplibre.org/maplibre-style-spec/layers/#text-variable-anchor) accepts an ordered list of candidate anchors and tries them sequentially.

korvin89:
🤖 AI generated

Closes #657

## What's added

Data label placement options for line series:

```ts
{
    type: 'line',
    name: 'Series',
    data: [...],
    dataLabels: {
        enabled: true,
        placement: 'auto', // or an exact order: ['bottom', 'top']
        placementFallback: 'show', // or 'hide'
    },
}
```

- `placement` — candidate positions relative to the point, tried in order: `'auto'` tries the built-in order (top, bottom, left, right), an array sets the exact order. Default: `['top']`.
- `placementFallback` — what to do when none of the positions fits: `show` draws the label at the first position anyway (crossing lines and labels is allowed then), `hide` hides the label. Default: `show`.

## How it works

There is a single placement engine and no separate modes — the previous "always above the point" behavior is just the default configuration (`['top']` + `show`), which produces pixel-identical output.

For every labeled point the positions are tried in order. The label box of every candidate is shifted into the plot area first; a position fits when the shifted box does not cross a visible line of any line series (segment × box test: cheap bbox reject, then Liang–Barsky clipping, with the box inflated by half of the line width) and does not overlap already placed labels (unless `allowOverlap`). The first fitting position wins; when none fits, `placementFallback` decides. Every placed label becomes an obstacle for the labels placed after it, and when labels of several series compete for the same spot, the series listed first in `series.data` wins.

With a single position and the `show` fallback the outcome is predetermined — the first position shifted into the plot — so the engine skips all checks: this is both the fast path and the exact reproduction of the previous behavior.

Interpolated lines (`monotone`, `cardinal`) are followed through the same curve factory the renderer uses, flattened into straight pieces, so the collision geometry matches what is drawn. Label sizes are measured once per series and cached, since every line layer places the labels of the preceding series again to use them as obstacles.

The feature is documented in the Line series guide (`docs/diplodoc/pages/guides/series-types/line.md`) with three live examples.


src/core/shapes/line/prepare-data.ts
```diff
@@ -156,5 +162,110 @@ export const prepareLineData = async (args: {
...
+
+        const ownData = new Map(acc.map((d) => [d.series.id, d]));
+
+        let segments: PlacementSegment[] = [];
+        let obstacles: PlacementRect[] = [];
+        let toPlace: Array<{points: PointData[]; series: PreparedLineSeries}> = labeled.map(
+            (d) => ({points: d.points, series: d.series}),
+        );
+
+        if (needSegments || needSimulation) {
```
kuzmadom, line 188: Architectural performance risk, but not necessarily a blocker: line series are prepared as separate layers, while each auto-placement series reprojects all line series and repeats placement for the preceding ones. This results in roughly O(S²·P) complexity. It should be acceptable for a small number of series, but may become expensive with HTML data labels.

src/core/shapes/line/auto-placement.ts
```diff
@@ -0,0 +1,374 @@
...
+    yMin: number;
+}
+
+/** Mirrors the d3 line `defined()` predicate from the line renderer */
+function isDrawnPoint(p: PointData): p is PointData & {x: number; y: number} {
+    return p.x !== null && p.y !== null && !p.hiddenInLine;
+}
+
+/** Segments of all visible lines in the layer: pairs of array-adjacent drawn points */
+export function getLineSegments(
```
kuzmadom, line 177: The collision geometry does not match the rendered line. Obstacles are built as straight segments between data points, while the renderer supports monotone and cardinal interpolation. For interpolated curves, the algorithm may miss actual intersections or reject valid label positions.

kuzmadom (commented):
It would be nice to update the documentation to cover the new data label placement options, though this is not a blocker.

korvin89:
🤖 AI generated

Documented in the Line series guide (`docs/diplodoc/pages/guides/series-types/line.md`, section “Data label placement”) with three live examples, plus a pointer from the general data labels guide.

src/core/shapes/line/prepare-data.ts
korvin89: 🤖 AI generated

Fair point. Added a per-series cache of measured label sizes (`WeakMap<PreparedLineSeries, Map<text, size>>`), so every label is measured once per render instead of once per layer — the DOM measurement of HTML labels was the expensive part of the re-simulation. The remaining O(S²·P) is pure geometry over already measured boxes.

src/core/shapes/line/auto-placement.ts
korvin89: 🤖 AI generated

Fixed: obstacles are now built by drawing the line through the same curve factory the renderer uses, into a context that flattens Bézier pieces into straight segments (8 per piece). Covered by unit tests and a screenshot test with a cardinal curve.

kuzmadom (approved):

PR #667, opened 2026-08-19
feat: add chart config declarations and JSON Schema

kuzmadom:
Adds tooling for generating and publishing standalone chart configuration artifacts:

  - chart-config.d.ts for Monaco’s TypeScript language service.
  - chart-config.schema.json for JSON validation and autocomplete.
  - Excludes callback-only options from the JSON Schema.
  - Validates generated artifacts during CI.
  - Documents how to consume both artifacts.


src/chart-config.ts
```diff
@@ -0,0 +1,13 @@
+import type {ChartData} from './core/types';
+
+type JsonPrimitive = boolean | null | number | string;
+
+type JsonValue = JsonPrimitive | JsonValue[] | {[key: string]: JsonValue};
```
korvin89: 🤖 **AI generated**

🔴 **blocker**

`JsonValue` isn't exported, and because of that `ts-json-schema-generator` can't derive a stable name for it — it falls back to a source-position id. As a result, 29 of the 87 definitions in the published schema come out looking like `ChartSeries<alias-src_chart-config.ts-100-176-src_chart-config.ts-0-502>`, plus there's a top-level definition literally named `alias-src_chart-config.ts-100-176-src_chart-config.ts-0-502` and 28 `$ref`s carrying `%3C…%3E`.

Two consequences worry me here. First, the source path and character offsets leak into a published artifact. Second, the names aren't stable: I appended a single unrelated comment line to `src/chart-config.ts` and every one of them shifted from `…-0-502` to `…-0-553`. So any edit to this file renames half the schema's definitions.

The existing guards don't catch it: `validateSchema` tests the raw `$ref` against `[<>\s]`, but the ref is URI-encoded, so `%3C` slips straight through. Once released, these names become part of the 1.x contract that `docs/diplodoc/pages/integrations/chart-config.md` promises to maintain.

I'd suggest one extra keyword — I verified this, and it brings the positional names down to 0, with definitions becoming `ChartSeries<JsonValue>`:

```ts
export type JsonValue = JsonPrimitive | JsonValue[] | {[key: string]: JsonValue};
```

src/chart-config.ts
```diff
@@ -0,0 +1,13 @@
...
+
+/**
+ * A chart configuration used to generate the published Monaco declaration bundle and JSON Schema.
+ *
+ * Callback functions remain available in TypeScript declarations. The artifact build omits them
+ * from JSON Schema because JSON cannot represent them.
+ */
+export interface ChartConfig extends ChartData<JsonValue> {}
```
korvin89, line 14: 🤖 **AI generated**

🟠 **should-fix**

It looks like `ChartData<JsonValue>` doesn't actually constrain `custom`: all the series interfaces are declared as `interface XSeries<T = MeaningfulAny> extends BaseSeries` — without `<T>` (see `src/core/types/chart/pie.ts:31`, `line.ts:58`, `bar-x.ts:40`, `area.ts:56` and the rest), so `BaseSeries.custom` falls back to `MeaningfulAny`.

You can see it in the generated schema: 16 `$ref: "#/definitions/MeaningfulAny"`, and the definition itself is `{"description": "A utilitarian type…"}` with no `type`, so it matches any value. That means `any` — and the internal helper name `MeaningfulAny` — end up in both published artifacts.

Maybe worth fixing in the types themselves, i.e. `extends BaseSeries<T>`?

scripts/build-chart-config.js
```diff
@@ -0,0 +1,599 @@
...
+    }
+
+    return [...errorsByIndex].map(([index, errors]) => ({
+        errors,
+        schema: defaultEntries[index].schema,
+        value: defaultEntries[index].value,
+    }));
+}
+
+function removeInvalidDefaults(schema) {
```
korvin89, line 494: 🤖 **AI generated**

🟠 **should-fix**

`removeInvalidDefaults` silently discards invalid `@default` tags. On the current sources that's 48 of them — for example `borderWidth: number` with `@default '1px'`, `boolean` fields with `@default 'series.dataLabels.enabled'`, and a `string` field with `@default "undefined"`.

Those look like genuine JSDoc bugs in `src/core/types/**`, which also feed the generated API docs — and the build now hides them. `getInvalidDefaults` already returns the full Ajv `errors` along with the schema and the value, but all of it is thrown away.

It would be great to at least log them (path + value), and ideally fail the build so the JSDoc gets fixed.

scripts/build-chart-config.artifact-test.js
```diff
@@ -0,0 +1,255 @@
...
+            /contains an invalid \$ref/,
+        ],
+    ])('rejects %s', (_testCase, mutateSchema, expectedError) => {
+        const invalidSchema = structuredClone(schema);
+        mutateSchema(invalidSchema);
+
+        expect(() => validateSchema(invalidSchema)).toThrow(expectedError);
+    });
+
+    test('validates the generated schema', () => {
```
korvin89, line 270: 🤖 **AI generated**

🟠 **should-fix**

What concerns me here is that there are no positive fixtures at all. The only "the schema accepts a valid config" assertion anywhere is `{series: {data: []}}` in `build-chart-config.js:503`; everything else is a negative case.

With `additionalProperties: false` everywhere, callback stripping, and default removal, an over-aggressive normalizer or a generator upgrade could quietly start rejecting real configs and CI wouldn't notice. On top of that, the artifacts aren't committed, so removing or renaming a config field produces no reviewable diff — even though the docs promise major-version discipline for exactly that kind of change.

I'd suggest adding:
1. a set of representative valid configs (line/bar/pie/area plus axes/tooltip/split), asserted to validate;
2. a committed golden snapshot of the definition and property name sets.

The second one only makes sense after the `JsonValue` fix, though — otherwise the snapshot would churn on every edit to that file.

scripts/build-chart-config.js
```diff
@@ -0,0 +1,599 @@
...
+
+function normalizeSchema(schema) {
+    normalizeSchemaNodes(schema, schema);
+    removeUnusedDefinitions(schema);
+    removeInvalidDefaults(schema);
+
+    const chartConfig = schema.definitions?.ChartConfig;
+
+    if (chartConfig) {
+        chartConfig.description =
```
korvin89: 🤖 **AI generated**

🟠 **should-fix**

The generated schema carries no `$id`, no `title`, and no version — the top level has exactly three keys: `$schema`, `$ref`, `definitions`.

Meanwhile the docs page states that the artifacts are versioned and that "the package version is also the schema version". So a consumer holding the file has no way to tell which version it is, and without `$id` the schema can't be referenced by URI — including from `monaco.languages.json.jsonDefaults`.

I'd suggest adding `$id`, `title`, and the package version right here, next to the existing `description` assignment.

scripts/build-chart-config.js
```diff
@@ -0,0 +1,599 @@
...
+                    noBanner: true,
+                },
+            },
+        ],
+        {preferredConfigPath: TSCONFIG_FILE},
+    );
+
+    validateDeclaration(DECLARATION_PATH, declaration);
+
+    const schema = createGenerator({
```
korvin89, line 614: 🤖 **AI generated**

🟠 **should-fix**

The pipeline strips callbacks but nothing else. The published schema still carries 2 occurrences of `"Not intended for public use"` (e.g. `ValueFormat.labelMode`) and 7 `deprecated` markers.

Shipping those in a stable JSON Schema effectively invites external configs to depend on them, and breaking that later would need a major bump.

Would it make sense to add an `@internal`/`@deprecated` exclusion pass mirroring the callback one — or, at minimum, to document explicitly that the schema isn't filtered on those markers?

scripts/build-chart-config.js
```diff
@@ -0,0 +1,599 @@
...
+        return;
+    }
+
+    visited.add(schema);
+
+    // Remove non-JSON properties before descending so discarded subtrees are not traversed.
+    // With functions: 'hide', ts-json-schema-generator represents callback-only properties as
+    // null or a closed empty object. Closed empty branches can also represent non-JSON class
+    // instances (for example Duration), so they are removed from unions below as well.
+    if (schema.properties) {
```
korvin89, line 318: 🤖 **AI generated**

🟠 **should-fix**

`normalizeSchemaNodes` inspects a node's `properties` before descending, so a definition that only becomes callback-only after its own normalization isn't detected in the nodes that reference it.

That makes the result depend on the key order in `definitions`: with two synthetic schemas differing only in key order, one dies with `contains a callback-only property` (and without naming the property), while the other normalizes correctly. It doesn't trigger on the current real types, but an order-of-keys dependency tends to surface eventually.

I'd suggest either running normalization to a fixpoint (repeat until nothing is removed), or normalizing `definitions` bottom-up.

scripts/build-chart-config.js
```diff
@@ -0,0 +1,599 @@
...
+                        `Cannot omit required callback-only property "${propertyName}" from chart-config.schema.json`,
+                    );
+                }
+
+                delete schema.properties[propertyName];
+            }
+        }
+    }
+
+    if (Array.isArray(schema.anyOf)) {
```
korvin89, line 333: 🤖 **AI generated**

🟠 **should-fix**

The `anyOf` filter can leave an empty array behind, which is a meta-schema violation. If a union consists only of closed-empty branches — say `renderer?: (() => void) | SomeClassInstance` — the property itself isn't removed, and the build fails with an opaque `Invalid chart config JSON Schema: … anyOf must NOT have fewer than 1 items` instead of the targeted callback-only message.

Failing closed is good, of course. But it's probably nicer to drop the property after filtering, exactly the way `isCallbackArtifact` already does.

korvin89 (commented):
🤖 **AI generated**

**Severity legend**

scripts/build-chart-config.js
```diff
@@ -0,0 +1,599 @@
...
+
+    if (validateConfig({series: {data: []}, unknownProperty: true})) {
+        throw new Error('chart-config.schema.json must reject unknown top-level properties');
+    }
+
+    if (validateConfig({series: {data: [], unknownProperty: true}})) {
+        throw new Error('chart-config.schema.json must reject unknown nested properties');
+    }
+
+    const invalidDefaults = getInvalidDefaults(schema);
```
korvin89: 🤖 **AI generated**

🟠 **should-fix**

Both failures here come out without any context, which will make them painful to debug.

Just above, `invalidDefaults` is computed with `schema`, `value`, and `errors` — and none of that reaches the message. In the check below, the loop iterates `Object.values(schemaNode.properties)`, so the offending property name isn't even bound; I'd suggest switching to `Object.entries` and putting the name (ideally the path) into the error text.

Separately: the `invalidDefaults` check is unreachable in the normal pipeline, since `normalizeSchema` → `removeInvalidDefaults` has already cleared them by that point. The same tautology applies to the assertion on line 221 of the test.

docs/diplodoc/pages/integrations/chart-config.md
```diff
@@ -0,0 +1,53 @@
+# Chart config tooling
+
+The package publishes two versioned config artifacts:
+
+- `@gravity-ui/charts/chart-config.d.ts` — standalone declarations for Monaco's TypeScript language service.
+- `@gravity-ui/charts/chart-config.schema.json` — JSON Schema for generation and validation of JSON-only chart configs.
+
+## Usage
+
+The declaration is a text asset, not a runtime JavaScript module. It must be loaded as text before it can be registered in Monaco. For example, Vite supports raw-file imports with the `?raw` suffix:
```
korvin89, line 12: 🤖 **AI generated**

🟠 **should-fix**

I don't think the Monaco example works as written. The generated bundle ends with `export {}`, which makes it a module — so `addExtraLib` alone isn't enough: `ChartConfig` won't be in the editor buffer's global scope.

I'd suggest either showing `import type {ChartConfig} from '@gravity-ui/charts/chart-config'` in the buffer itself, or recommending a `declare global` wrapper.

One more request: the page's headline promises "JSON validation and autocomplete", but there's no `monaco.languages.json.jsonDefaults.setDiagnosticsOptions({schemas: [...]})` example — only the TypeScript half is documented. It would be great to cover the JSON side too.

src/index.ts
```diff
@@ -1,4 +1,5 @@
 export {CustomShapeRenderer} from './utils';
+export type {ChartConfig} from './chart-config';
```
korvin89: 🤖 **AI generated**

🟠 **should-fix**

This export makes `ChartConfig` a permanent, semver-stable root export and a second, subtly different name for `ChartData` — while the component's actual prop is `data: ChartData`. There's a real risk people will take it for the canonical config type.

The generator doesn't need the re-export either: it points at `ENTRY_FILE = src/chart-config.ts` directly.

Would it work to keep `src/chart-config.ts` as a build-only entry point? Or, if a public type really is wanted, to give it a name that says what it is — e.g. `JsonChartConfig`.

src/chart-config.ts
```diff
@@ -0,0 +1,13 @@
+import type {ChartData} from './core/types';
+
+type JsonPrimitive = boolean | null | number | string;
+
+type JsonValue = JsonPrimitive | JsonValue[] | {[key: string]: JsonValue};
+
+/**
```
korvin89, line 7: 🤖 **AI generated**

🔵 **nit**

This ends up as two sources of truth for one description: this JSDoc lands in the `.d.ts` and is what shows on hover in Monaco, while the schema gets the hard-coded string from `normalizeSchema`, which overrides it. So the two artifacts describe the same type differently.

Also, the text itself is about the build ("used to generate the published Monaco declaration bundle…") rather than about what a consumer gets. Maybe worth rewriting it from the user's point of view and reusing it in both artifacts?

scripts/build-chart-config.js
```diff
@@ -0,0 +1,599 @@
...
+
+        ts.forEachChild(node, visit);
+    }
+
+    visit(sourceFile);
+
+    return [...references];
+}
+
+function validateDeclaration(filePath, declaration) {
```
korvin89, line 68: 🤖 **AI generated**

🔵 **nit**

`getExternalDeclarationReferences(declaration, filePath)` and `validateDeclaration(filePath, declaration)` take the same two arguments in the opposite order. Both are exported and both params are strings, so a swapped call won't fail — it'll just quietly misbehave. I'd suggest aligning the order.

scripts/build-chart-config.js
```diff
@@ -0,0 +1,599 @@
...
+function getDefinitionName(reference) {
+    const prefix = '#/definitions/';
+
+    if (typeof reference !== 'string' || !reference.startsWith(prefix)) {
+        return undefined;
+    }
+
+    const encodedName = reference.slice(prefix.length).split('/')[0];
+
+    return decodeURIComponent(encodedName).replaceAll('~1', '/').replaceAll('~0', '~');
```
korvin89: 🤖 **AI generated**

🔵 **nit**

`decodeURIComponent` throws `URIError` on a bare `%` in a definition name, so the build would die with an opaque stack instead of a readable message. A `try/catch` returning `undefined` would be enough.

scripts/build-chart-config.artifact-test.js
```diff
@@ -0,0 +1,255 @@
...
+    let schema;
+
+    beforeAll(() => {
+        ({declaration, schema} = generateChartConfigArtifacts());
+    });
+
+    test('generates a standalone declaration bundle', () => {
+        expect(getExternalDeclarationReferences(declaration, DECLARATION_PATH)).toEqual([]);
+        expect(declaration).toContain('export interface ChartConfig');
+        expect(Buffer.byteLength(declaration)).toBeLessThan(150_000);
```
korvin89, line 39: 🤖 **AI generated**

🔵 **nit**

`toBeLessThan(150_000)` reads as a bit of a mystery without a comment — the bundle is currently 89,527 bytes. It's not clear from the test what the threshold guards against (accidental library inlining?), and it'll fire confusingly the day the type surface legitimately grows.

I'd suggest a one-line comment explaining the intent — or dropping the check, since the real invariant is already covered by the "no external references" assertion.

scripts/build-chart-config.artifact-test.js
```diff
@@ -0,0 +1,255 @@
...
+    });
+
+    test('validates declaration content without accessing the published file', () => {
```
korvin89, line 42: 🤖 **AI generated**

🔵 **nit**

A bare `.toThrow()` with no matcher: the test passes even if `validateDeclaration` fails for a completely unrelated reason — missing tsconfig, bad compiler options, and so on. I'd suggest asserting on the message, e.g. `/MissingType|TS2304/`.

And a related note on line 18: `DECLARATION_PATH` points at `scripts/chart-config.d.ts`, a path that never exists, and it only works because the compiler host is overridden. A one-line comment would help the next reader a lot.

jest.chart-config.config.js
```diff
@@ -0,0 +1,8 @@
+const baseConfig = require('./jest.config');
+
+module.exports = {
+    ...baseConfig,
+    setupFiles: [],
+    testMatch: ['<rootDir>/scripts/**/*.artifact-test.js'],
+    testPathIgnorePatterns: [],
```
korvin89: 🤖 **AI generated**

🔵 **nit**

`testPathIgnorePatterns: []` drops Jest's default `/node_modules/` guard. It's safe today only because `testMatch` is scoped to `<rootDir>/scripts/**` — `['/node_modules/']` would express the intent more precisely.

A related observation: `npm run test` doesn't pick these up (the main run has 16 suites, none of them `*.artifact-test.js`), so a local `npm test` gives zero coverage of the generator. If that isn't deliberate, Jest `projects` would let one command run both sets.

package.json
```diff
@@ -8,6 +8,8 @@
   "types": "./dist/esm/index.d.ts",
   "exports": {
     "./package.json": "./package.json",
+    "./chart-config.d.ts": "./dist/config/chart-config.d.ts",
+    "./chart-config.schema.json": "./dist/config/chart-config.schema.json",
     ".": {
       "import": {
         "types": "./dist/esm/index.d.js",
```
korvin89: 🤖 **AI generated**

🔵 **nit**

This one is pre-existing rather than introduced by the PR, but since the `exports` block is being touched here: `"types": "./dist/esm/index.d.js"` should be `.d.ts`. It's a one-character fix.

.github/workflows/ci.yml
```diff
@@ -45,3 +45,5 @@ jobs:
         run: npm ci
       - name: Unit Tests
         run: npm run test
+      - name: Test Chart Config Artifacts
```
korvin89, line 48: 🤖 **AI generated**

🔵 **nit**

Slight duplication of work: the `verify_files` job already runs the full generator along with `validateDeclaration` and `validateSchema` via `npm run build`, and this step regenerates everything again (roughly 4–7 seconds).

So it's belt-and-braces rather than new coverage — quite possibly on purpose. Just flagging it so it's explicit.

scripts/build-chart-config.js
```diff
@@ -0,0 +1,599 @@
...
+
+    if (definitionName) {
+        references.add(definitionName);
+    }
+
+    for (const childSchema of getSchemaChildren(schema)) {
+        if (
+            skipDefinitions &&
+            schema.definitions &&
+            Object.values(schema.definitions).includes(childSchema)
```
korvin89: 🤖 **AI generated**

🔵 **nit**

A couple of small things around performance and noise:

- `Object.values(schema.definitions).includes(childSchema)` is recomputed for every child, which makes this O(n²) over 87 definitions — building a `Set` of the definition objects once would do.
- In `writeChartConfigArtifacts` (lines 570–571) there are two `mkdirSync` calls for what is, in practice, always the same directory.

src/chart-config.ts
```diff
@@ -0,0 +1,13 @@
+import type {ChartData} from './core/types';
```
korvin89: 🤖 **AI generated**

🔵 **nit**

`import type {ChartData} from './core/types'` — the other files at the root of `src/` go through the `~core/*` alias (see `src/index.ts:2-4`). I'd suggest `~core/types` for consistency.

src/chart-config.ts
kuzmadom: Deferring the fix. Threading <T> narrows a public type and would break any external consumer relying on the current shape, so it belongs in the next major with a changelog entry rather than in this PR or a near-term follow-up.

scripts/build-chart-config.js
kuzmadom: The schema is versioned with the package, deprecated in the schema = deprecated in the package, this is the contract

korvin89 (approved):

PR #668, opened 2026-08-24
feat: add interpolation option for line series

### LINKED ISSUE #658: Add configurable interpolation for line series
## Problem

  Line-series points are currently connected with straight segments. Users may want a visually smoother line while keeping all data points at their original coordinates.

  The existing linejoin option only changes the appearance of segment joins and does not interpolate the path.

  ## Proposal

  Add an opt-in interpolation option to LineSeries:

```
  {
      type: 'line',
      interpolation: {
          type: 'monotone',
      },
  }
```

  When the option is omitted, the current linear interpolation is used.

  The object form leaves room for additional interpolation types and type-specific options in the future:

```
  interpolation: {
      type: 'cardinal',
      tension: 0.5,
  }
```

  ## Expected behavior

  With type: 'monotone':

  - The curve passes through every original data point.
  - No artificial extrema or overshoots are introduced between adjacent points.
  - Markers, tooltips and interactions remain attached to the original points.
  - Existing nullMode behavior is preserved.
  - No trend line, moving average or additional data values are calculated.

  ## Scope

  The option should be declared on LineSeries, not BaseSeries, and implemented inside the line plugin.

  Area-series support can be considered separately.

  The project already uses d3-shape, which provides curveMonotoneX (https://d3js.org/d3-shape/curve) for this behavior.

  Similar interpolation settings are available in Power BI (https://learn.microsoft.com/en-gb/power-bi/visuals/power-bi-line-chart), Apache ECharts (https://echarts.apache.org/handbook/en/how-to/chart-types/line/smooth-line/), Chart.js (https://www.chartjs.org/docs/latest/charts/line.html), and Vega (https://vega.github.io/vega/docs/marks/line/).

kuzmadom:
## Summary                                                                                                            
   
  - Added `interpolation` option to `LineSeries` with three types: `linear`, `monotone`, `cardinal` (with optional `tension`)                                                
  - Implemented via d3-shape curves (`curveMonotoneX`, `curveCardinal`) in the line renderer                            
  - Markers, tooltips, and interactions remain attached to original data points                                         
  - `nullMode` behavior preserved            

Closes #658


src/plugins/line/index.ts
```diff
@@ -13,10 +14,47 @@ import {getTooltipLineSymbol} from '~core/tooltip/utils';
 import {filterLayerLabels} from '~core/utils';
 import {validateAxisPlotValues, validateXYSeries} from '~core/validation/helpers';
 
+import {CHART_ERROR_CODE, ChartError} from '../../libs';
 import type {LineSeries} from '../../types';
 
 import {prepareLineSeries} from './prepare-line-series';
 
+const AVAILABLE_INTERPOLATION_TYPES = ['linear', 'monotone', 'cardinal'];
```
korvin89: 🤖 **AI generated**

🟠 **should-fix**

This validation sits apart from where the rest of the series validation lives. `src/core/validation/helpers.ts:284` already has almost exactly this shape — `validateStacking({series}: {series: {stacking?: string}})`, using the same `CHART_ERROR_CODE.INVALID_DATA` and the same `label_invalid-series-property` i18n key with a `key` + `values` array. All seven plugins keep their `validate:` down to composing helpers from `~core/validation/helpers`, whereas this file now defines `AVAILABLE_INTERPOLATION_TYPES`, `throwInvalidInterpolation` and `validateInterpolation` locally, plus a direct `import {CHART_ERROR_CODE, ChartError} from '../../libs'` mixed in with the `~core/*` imports.

Would you consider moving `validateInterpolation` into `helpers.ts` next to `validateStacking`, and aligning its signature with it (`{series}: {series: {interpolation?: unknown}}` rather than taking the series object itself)? That would keep all series-level validation in one place.

src/core/shapes/line/renderer.ts
```diff
@@ -28,17 +28,31 @@ export function renderLine(
...
-        .x((d) => d.x as number)
-        .y((d) => d.y as number);
-
     plotSvgElement.selectAll('*').remove();
     const lineSelection = plotSvgElement
         .selectAll('path')
         .data(preparedData)
         .join('path')
-        .attr('d', (d) => line(d.points))
+        .attr('d', (d) => {
```
korvin89: 🤖 **AI generated**

🔵 **nit**

The curve resolution is inlined into the `.attr('d')` accessor, and the `lineGenerator` (with its three accessor closures) moved in with it — it used to be built once per render and is now rebuilt for every series.

The repo already has this factored out for the pie shape: `getCurveFactory(data): CurveFactory` in `src/core/shapes/pie/utils.ts:12`. It might be nice to mirror that — a `getCurveFactory(interpolation)` next to the line shape, the generator hoisted back out, and `.curve(...)` applied per datum. Purely a readability/consistency suggestion, up to you.

docs/diplodoc/pages/guides/line-interpolation.md
```diff
@@ -0,0 +1,36 @@
...
+    interpolation: {type: 'cardinal', tension: 0.5},
+}
+```
+
+<div data-chart-example="line-interpolation/cardinal"></div>
+
+## Notes
+
+- Curves pass through every original data point.
+- `monotone` requires data points to be ordered monotonically by their x values.
```
korvin89, line 34: 🤖 **AI generated**

🔵 **nit**

Good call adding the `monotone` ordering note. One more caveat might be worth having here: with `cardinal` and `tension < 1` the curve overshoots the data's min/max between points, and since the series group is clipped to the plot rect (`src/components/ChartInner/index.tsx:333`, `src/hooks/useShapes/SeriesShapes.tsx:71`) that overshoot gets cut flat at the plot edge when a point sits at the axis extreme.

That's really the deciding factor between the two types for someone choosing one, so it would be great to see it spelled out next to "Curves pass through every original data point".

korvin89:
🤖 **AI generated**

**Severity legend**

src/plugins/line/index.ts
kuzmadom: Moved the validation to `src/plugins/line/validation.ts` and aligned its API. I kept it in the line plugin because interpolation is a line-specific feature.

src/core/shapes/line/renderer.ts
kuzmadom: done

docs/diplodoc/pages/guides/line-interpolation.md
kuzmadom: Added a note about cardinal overshoot and clipping at the plot boundary.

korvin89 (approved):

PR #669, opened 2026-08-26
feat: add gradient colors for line and area series

kuzmadom:
## Summary

  Add configurable linear gradients for line and area series, with independent stroke and fill colors for areas.

  - Render gradients in line and area charts
  - Use the gradient midpoint color in legends
  - Resolve gradient colors at point positions for tooltips and hover markers
  - Brighten gradient colors on hover
  - Validate gradient configuration
  - Add Storybook examples


src/core/shapes/line/prepare-data.ts
```diff
@@ -99,6 +100,21 @@ export const prepareLineData = async (args: {
...
+            if (point.color !== undefined) {
+                return point.color;
+            }
+            return s.gradient && gradientBBox && point.x !== null && point.y !== null
+                ? getGradientColorAtPoint(point.x, point.y, s.gradient, gradientBBox)
+                : s.color;
+        };
+
+        for (const point of points) {
+            point.color = getMarkerFill(point);
```
korvin89: 🤖 **AI generated**

🟠 **should-fix**

I'd suggest not overwriting `point.color` here. The field currently carries what the user configured (`d.marker?.color ?? d.color`), and this loop replaces it with a derived value, so the "user set it" / "we computed it" distinction is lost. Storing the resolved value in a separate field (e.g. `point.fill`) and leaving `point.color` as the raw input would keep both.

It would also untangle some now-redundant plumbing: once every point has a `color`, `fill: getMarkerFill(p)` on line 131 always takes the first branch, the third argument of `buildHoverMarkerGetter(points, s, getMarkerFill)` becomes equivalent to its default, and `p.color ?? p.series.color` in `get-tooltip-data.ts` never reaches the fallback. Right now two mechanisms encode the same rule — it'd be great to keep just one.

One more small thing: for series without a gradient the loop only rewrites `undefined → s.color` with no observable effect, so guarding it with `if (s.gradient)` would save the pass.

The same applies to `src/core/shapes/area/prepare-data.ts:366`.

src/core/shapes/line/renderer.ts
```diff
@@ -9,13 +9,53 @@ import type {LabelData, TooltipDataChunkLine} from '../../../types';
...
 const b = block('line');
 
+interface LineGradientIds {
+    hover?: string;
+    normal?: string;
+}
+
+const gradientIds = new WeakMap<PreparedLineData, LineGradientIds>();
+
+function strokeRef(
```
korvin89: 🤖 **AI generated**

🟠 **should-fix**

`strokeRef` here and `lineStrokeRef` / `areaFillRef` in `src/core/shapes/area/renderer.ts:44,77` are three near-verbatim copies of the same "bbox → WeakMap cache → `ensureGradientDef` → `url(#id)`" sequence, backed by two separate module-level `gradientIds` maps. The two `getMarkerFill` closures in the prepare-data files are a fourth copy of the resolution rule.

What makes me raise it above style: the "clear the cache before re-rendering" invariant is duplicated too (`line/renderer.ts:77`, `area/renderer.ts:129`), and it's load-bearing — a stale id produces `stroke="url(#missing)"`, which has no fallback and simply doesn't render. A third shape type adopting gradients would have to rediscover that.

Would it be worth extracting a single `createGradientRef(plot)` resolver (and a shared marker-fill resolver) into `core/utils/gradient.ts`?

src/core/utils/gradient.ts
```diff
@@ -0,0 +1,184 @@
...
+): string {
+    if (!container.ownerSVGElement) {
+        return '';
+    }
+    const containerSelection = select(container);
+    let defs = containerSelection.select<SVGDefsElement>('defs.gradients');
+    if (defs.empty()) {
+        defs = containerSelection.append('defs').attr('class', 'gradients');
+    }
+    const id = `chart-gradient-${gradientIdCounter++}`;
```
korvin89: 🤖 **AI generated**

🟠 **should-fix**

The generated ids are only unique within one module instance. If two copies of the library end up on the same page — a duplicated version in the npm tree, or a micro-frontend setup — both counters start at 0 and the document ends up with two `id="chart-gradient-0"`. `url(#chart-gradient-0)` resolves to whichever comes first in document order, so one chart would silently repaint with another chart's gradient, with no error anywhere.

Could we mix in something already unique? `PreparedLineData` / `PreparedAreaData` carry an `id` from `getUniqId()`, so something like `${data.id}-gradient-line-normal` would be collision-free and easier to recognise while debugging.

src/core/series/prepare-legend.ts
```diff
@@ -130,6 +130,7 @@ function getFlattenLegendItems(series: PreparedSeries[], preparedLegend: Prepare
         if (s) {
             acc.push({
                 ...s,
+                color: s.legend.color ?? s.color,
```
korvin89, line 133: 🤖 **AI generated**

❓ **question**

The PR description says gradients are rendered "in charts, legends, tooltips, and hover markers", but the legend only ever receives the flat `getGradientMidColor` swatch — which the new guides state explicitly ("it does not render a gradient symbol"). Would it be worth adjusting the wording in the PR body / release note, or is a gradient legend symbol planned as a follow-up?

src/core/utils/gradient.ts
```diff
@@ -0,0 +1,184 @@
...
+    return interpolateGradientColor(t, gradient.stops);
+}
+
+let gradientIdCounter = 0;
+
+/**
+ * Ensures a <linearGradient> def exists in the container's <defs> and returns its id.
+ * Uses gradientUnits="userSpaceOnUse" with coordinates derived from the bbox and angle.
+ */
+export function ensureGradientDef(
```
korvin89: 🤖 **AI generated**

🔵 **nit**

The name promises idempotence that isn't here: the function unconditionally appends a new `<linearGradient>` and returns a fresh id — only the `<defs>` container is actually "ensured", and the real caching lives in the callers' WeakMaps. `createGradientDef` (or `appendGradientDef`) would describe it more accurately.

src/core/utils/gradient.ts
```diff
@@ -0,0 +1,184 @@
...
+    if (len2 === 0) {
+        // SVG renders a degenerate linear gradient using the last stop color.
+        return 1;
+    }
+
+    return Math.max(0, Math.min(1, ((px - coords.x1) * dx + (py - coords.y1) * dy) / len2));
+}
+
+function interpolateGradientColor(t: number, stops: GradientStop[]): string {
+    const sorted = [...stops].sort((a, b) => a.offset - b.offset);
```
korvin89: 🤖 **AI generated**

🔵 **nit**

`interpolateGradientColor` copies and re-sorts the stops array on every call, and `getGradientColorAtPoint` (line 146) recomputes `gradientAngleToCoords` for each point. Both are called once per data point in the prepare loop, so the sorted stops and the coords could be hoisted out of the per-point path.

src/plugins/area/prepare-area-series.ts
```diff
@@ -68,7 +68,13 @@ export function prepareAreaSeries(args: PrepareSeriesArgs<AreaSeries>) {
     return seriesList.map<PreparedAreaSeries>((series) => {
         const id = getUniqId();
         const name = series.name || '';
-        const color = series.color || colorScale(name);
+        const lineColor = series.color || colorScale(name);
+        const areaFillColor = series.fillColor ?? lineColor;
```
korvin89: 🤖 **AI generated**

🔵 **nit**

Small asymmetry with the line above: `series.color || colorScale(name)` treats `''` as "not set", while `series.fillColor ?? lineColor` keeps it. I checked locally — `fillColor: ''` ends up as `fillColor === ''` and `legend.color === ''`, which becomes `.attr('fill', '')` on the region (an invalid paint, so the browser falls back to black) and an empty color in the legend. Since `color: ''` deliberately falls back to the palette (there's an explicit test for it in `validation.test.ts`) and the validator accepts `''`, using `||` here would keep the two consistent.

src/core/types/chart/gradient.ts
```diff
@@ -0,0 +1,35 @@
...
+    color: string;
+    /**
+     * Position of the stop as a value from 0 to 1.
+     * @minimum 0
+     * @maximum 1
+     */
+    offset: number;
+}
+
+export interface LinearGradient {
```
korvin89, line 15: 🤖 **AI generated**

🔵 **nit**

The neighbouring public types in this folder keep their fields alphabetical (`Halo`: `enabled, opacity, size`; `PointMarkerOptions`: `borderColor, borderWidth, color, enabled, radius, symbol`), and the generated schema snapshot lists `["angle", "stops", "type"]`. `LinearGradient` is `type, angle, stops` — worth sorting for consistency (unless `type` is intentionally kept first as the discriminant).

Same thing for the new `color?` in `src/core/types/chart/tooltip.ts:55` and `:68`, which is placed after `data`.

src/core/utils/__tests__/gradient.test.ts
```diff
@@ -0,0 +1,104 @@
...
+    getBrighterGradient,
+    getGradientColorAtPoint,
+    getGradientMidColor,
+    gradientAngleToCoords,
+    isLinearGradient,
+} from '../gradient';
+
+const BBOX = {xMin: 0, xMax: 100, yMin: 0, yMax: 50};
+
+describe('gradientAngleToCoords', () => {
```
korvin89, line 13: 🤖 **AI generated**

🔵 **nit**

This block covers `0` and `270` only. It might be worth adding `180` — the default every gradient without an explicit `angle` uses — and `90`, since both rely on the `TRIGONOMETRIC_EPSILON` snap to come out axis-aligned. A non-axis-aligned angle would also be nice, as that's the only case where the `|hw·sin| + |hh·cos|` length formula actually does anything.

src/core/types/chart/gradient.ts
```diff
@@ -0,0 +1,35 @@
+/* eslint-disable jsdoc/check-tag-names */
```
korvin89: 🤖 **AI generated**

🔵 **nit**

This file-level disable is needed only because `@minimum` / `@maximum` / `@minItems` / `@minLength` (consumed by `ts-json-schema-generator`) aren't registered with the eslint jsdoc plugin. Registering them once in `settings.jsdoc.definedTags` in `eslint.config.mjs` would save repeating the disable in every future type that carries schema annotations.

korvin89:
🤖 **AI generated**

**Severity legend**

src/core/series/prepare-legend.ts
kuzmadom: Updated the PR description. The legend uses the gradient midpoint color.

korvin89 (approved):

PR #671, opened 2026-08-27
fix(pie): recalculate data label percentage for visible slices

### LINKED ISSUE #670: Pie chart: data label percentage is not recalculated when slices are hidden via `visible: false`
When pie slice visibility changes (e.g. legend click with modifier key), **slice angles are recalculated correctly**, but **data label text keeps the original percentage** from the initial render.

This is a regression compared to Highcharts behavior (`{point.percentage}`), which dynamically reflects the share among currently visible slices.

---

## Steps to reproduce

1. Create a pie chart with 3 slices, e.g. values `10`, `10`, `5` (shares: 40%, 40%, 20%).
2. Enable data labels (`dataLabels.enabled: true`).
3. Set labels with percentage, e.g. via `point.label`: `"10.00 (40.00%)"`.
4. Hide one slice by setting `visible: false` on a point (or toggling via legend).
5. Re-render the chart with updated config.

---

## Expected behavior

- Visible slices still sum to 100% in labels.
- After hiding the `5` slice, the two remaining slices show **50%** each.
- Label percentages match slice angles.

---

## Actual behavior

- Slice geometry updates correctly (pie stays full circle, angles redistribute).
- Label text stays unchanged (still 40%, 40%, 20%).
- Percentages in labels no longer sum to 100%.

---

## Environment

- `@gravity-ui/charts`: 1.57.3 (via `@gravity-ui/chartkit` 8.0.1)
- Chart type: `pie`
- Data labels: `enabled: true`, `html: true`

---

Image

catakot:
Expose slice share in custom formatters and derive label text from segment angles when dataLabels.format is set, so percentages stay in sync when slices are hidden.

Fixes #670


kuzmadom:
Thank you for the contribution. After reviewing the proposed approach, we have decided not to merge this PR.

The implementation introduces Pie-specific fields into the shared formatter context and limits percentage support to Pie data labels. In the current architecture, series-specific formatter capabilities should be defined by the corresponding series plugins and applied consistently to both data label and tooltip formatters.

We are addressing the underlying issue in [#672](https://github.com/gravity-ui/charts/pull/672), which introduces typed per-series formatter contexts and extends percentage support to Pie and percent-stacked series. We will therefore continue with that implementation and close this PR.

PR #672, opened 2026-08-27
feat: expose percentage in data label and tooltip formatters

### LINKED ISSUE #670: Pie chart: data label percentage is not recalculated when slices are hidden via `visible: false`

kuzmadom:
Closes #670

  ## What changed

  Custom series data-label and tooltip formatters can now access computed percentages:

  - Pie: `{value, percentage, name, data}`.
  - Area, BarX and BarY: `{value, percentage, data}`, with percentages available for percent stacking.

  Additional context fields are optional. `percentage` represents a fraction in the `0..1` range, calculated from currently visible data and updated when legend visibility changes.

  ```ts
  formatter: ({value, percentage}) =>
      percentage === undefined
          ? String(value)
          : `${value} (${Math.round(percentage * 100)}%)`;
  ```

  Percentages use source values rather than pixel dimensions. For Area, isolated points between explicit nulls still count toward percentage totals, even when gaps prevent them from contributing to the filled area.

  Existing value-only formatters and helpers accepting `BaseSeries` remain compatible.

## Fixed

  - Percent-stacked Area, BarX and BarY now reject negative values with `INVALID_DATA`, preventing incorrect  normalization and rendering. Previously, these values passed validation. This includes negative numeric strings on   numeric value axes.
  - Zero-total percent stacks avoid division by zero.
  - Area percent-stack geometry correctly handles isolated points between explicit nulls.


korvin89 (commented):
🤖 **AI generated**

**Severity legend**

src/core/types/chart/pie.ts
```diff
@@ -28,7 +36,15 @@ export interface PieSeriesData<T = MeaningfulAny> extends BaseSeriesData<T> {
...
-export interface PieSeries<T = MeaningfulAny> extends BaseSeries {
+export interface PieFormatContext<T = MeaningfulAny>
+    extends CustomFormatContext, PercentageFormatContext {
+    name: string;
+    data: PieSeriesData<T>;
+}
+
+export type PieValueFormat<T = MeaningfulAny> = ValueFormat<PieFormatContext<T>>;
+
+export interface PieSeries<T = MeaningfulAny> extends Omit<BaseSeries, 'dataLabels' | 'tooltip'> {
```
korvin89: 🤖 **AI generated**

❓ **question** — Are we sure we want breaking changes in this PR? I can see two, and the PR is still titled `feat:` with no `BREAKING CHANGE` note, so semantic-release would ship them in a minor version:

1. **Types:** `PieSeries`, `AreaSeries`, `BarXSeries`, and `BarYSeries` now extend `Omit<BaseSeries, 'dataLabels' | 'tooltip'>`. As a result, none of them can be assigned to `BaseSeries` anymore, even when they don't use a series-specific formatter. For example, `function applyDefaults(s: BaseSeries)` compiles on `main` but fails here for all four types. I see `FunnelSeries` already uses this pattern, so maybe it's acceptable — I'd just like to confirm it's intentional.
2. **Runtime:** in `src/core/validation/helpers.ts`, negative values in a percent stack now throw `INVALID_DATA`. Charts that used to render (even if incorrectly) will show an error after the upgrade.

If both are intended, could we mark them as breaking? If not, we could keep `BaseSeries` compatibility and log a warning and clamp negatives instead of throwing.

src/core/validation/helpers.ts
```diff
@@ -6,13 +6,42 @@ import {DEFAULT_AXIS_TYPE} from '../constants';
...
+export function validatePercentStackingValues(args: {
+    series: PercentStackingValidationSeries;
+    valueKey: keyof XYDataPoint;
+}) {
+    const {series, valueKey} = args;
+    if (
+        series.stacking === 'percent' &&
+        series.data.some((point) => {
+            const value = point[valueKey];
+            return typeof value === 'number' && value < 0;
```
korvin89, line 38: 🤖 **AI generated**

🟠 **should-fix** — This check only catches numeric negatives (`typeof value === 'number' && value < 0`), but the public types allow strings here (`AreaSeriesData['y']` and `BarXSeriesData['y']` are `string | number | null`, and `BarYSeriesData['x']` is `string | number`). A value like `'-5'` passes validation and still reaches the percent normalization this check is meant to protect. Maybe something like this:

```ts
const value = Number(point[valueKey]);
return Number.isFinite(value) && value < 0;
```

src/core/shapes/area/prepare-data.ts
```diff
@@ -118,43 +117,55 @@ export const prepareAreaData = async (args: {
...
+                        }
+                        const data = seriesDataMaps.get(s);
+                        if (!data) {
+                            return;
+                        }
+                        const value = Number(data.get(x)?.y ?? 0);
+                        if (!Number.isFinite(value)) {
+                            return;
+                        }
+                        percentageTotal += Math.max(0, value);
```
korvin89, line 159: 🤖 **AI generated**

❓ **question** — Splitting `stackTotal` from `percentageTotal` fixes the geometry nicely, thank you! One thing I'd like to check is the semantics. Take a point between two explicit `null`s: it isn't drawn, but it still counts toward `percentageTotal`. So at that x the top series fills 100% of the height, yet its `percentage` is `0.5` (the test in `area/__tests__/prepare-data.test.ts:80` expects `[1, 0.5, 1]`). Its data label would read "10 (50%)" on a band that visually covers the whole plot. Is that the intended behavior? If so, could we mention it in the docs so users aren't surprised?

src/core/shapes/bar-x/__tests__/prepare-data.test.ts
```diff
@@ -0,0 +1,41 @@
...
+                series,
+                seriesOptions: seriesOptionsDefaults,
+                xAxis: {type: 'linear'} as PreparedXAxis,
+                xScale: scaleLinear().domain([0, 2]).range([0, 400]),
+                yAxis: [{type: 'linear', plotIndex: 0}] as PreparedYAxis[],
+                yScale: [scaleLinear().domain([0, 100]).range([height, 0])],
+                boundsHeight: height,
+                split: {plots: [{top: 0, height}]} as PreparedSplit,
+            });
+            expect(result.map((item) => item.percentage)).toEqual([0.25, 0.75]);
```
korvin89, line 42: 🤖 **AI generated**

🔵 **nit** — Thanks for adding these unit tests! Two small coverage gaps are left: bar-y has no unit test for the percentage calculation, and nothing checks that `percentage` is `undefined` when `stacking` isn't `'percent'` (the docs promise this). Also, `expect([...]).toEqual([true, true, true, true])` in `src/core/types/__tests__/series-format-context.test.ts:51` always passes, because `isBaseSeries` returns `true` unconditionally. Perhaps we could remove it, since the new `@ts-expect-error` checks already cover the typing.

src/core/shapes/bar-x/prepare-data.ts
```diff
@@ -340,8 +341,15 @@ export const prepareBarXData = async (args: {
                         (sum, item) => sum + item._height,
                         0,
                     );
-                    const ratio = currentPlotHeight / positiveStackHeight;
+                    const ratio =
+                        positiveStackHeight > 0 ? currentPlotHeight / positiveStackHeight : 0;
                     stackItems.forEach((item) => {
+                        item.percentage =
+                            item.series.stacking === 'percent' && positiveStackSum > 0
```
korvin89: 🤖 **AI generated**

🔵 **nit** — This PR adds three `no-nested-ternary` ESLint warnings for the same calculation: here, in `src/core/shapes/area/prepare-data.ts:220`, and in `src/core/shapes/bar-y/prepare-data.ts:180`. Could we move it into a shared helper? For example:

```ts
function getStackPercentage(args: {stacking?: string; value: number; total: number}) {
    const {stacking, value, total} = args;
    if (stacking !== 'percent') {
        return undefined;
    }
    return total > 0 && value > 0 ? value / total : 0;
}
```

src/core/types/chart/base.ts
```diff
@@ -12,10 +12,19 @@ type DateFormat = {
     format?: string;
 };
 
-export type CustomFormat = {
+export interface CustomFormatContext {
+    value: unknown;
+}
+
+export interface PercentageFormatContext {
```
korvin89, line 19: 🤖 **AI generated**

🔵 **nit** — `PercentageFormatContext` is exported but only used by `PieFormatContext`. `AreaFormatContext`, `BarXFormatContext`, and `BarYFormatContext` each redeclare `percentage?: number` with a copied JSDoc. Now that the field is optional everywhere, maybe they could all extend `PercentageFormatContext`? Similarly, `getValueFormatContext` is the same `(item) => ({percentage: item.percentage, data: item.data})` in the area, bar-x, and bar-y plugins and could be shared.

scripts/__snapshots__/chart-config.schema.summary.json
```diff
@@ -89,6 +93,10 @@
     "TreemapSeries<JsonValue>",
     "TreemapSeriesData<JsonValue>",
     "ValueFormat",
+    "ValueFormat<AreaFormatContext<JsonValue>>",
```
korvin89, line 96: 🤖 **AI generated**

🔵 **nit** — The public JSON schema now includes `ValueFormat<AreaFormatContext<JsonValue>>`, `ValueFormat<BarXFormatContext<JsonValue>>`, `ValueFormat<BarYFormatContext<JsonValue>>`, and `ValueFormat<PieFormatContext<JsonValue>>`. When I built it locally, all four were byte-for-byte identical to plain `ValueFormat` (about 3.5 KB each). `formatter` is a function, so a custom formatter can't be expressed in a JSON config anyway. Maybe these series-specific aliases could be left out of the config declaration?

src/plugins/bar-y/prepare-bar-y-series.ts
```diff
@@ -28,9 +28,16 @@ async function prepareDataLabels(series: BarYSeries) {
...
-    const labels = enabled
-        ? series.data.map((d) => getFormattedValue({value: d.label ?? d.x, ...series.dataLabels}))
-        : [];
+    const labels =
+        enabled && series.stacking !== 'percent'
+            ? series.data.map((d) =>
+                  getFormattedValue({
+                      value: d.label ?? d.x,
+                      format: series.dataLabels?.format,
+                      context: {data: d, percentage: undefined},
```
korvin89: 🤖 **AI generated**

🔵 **nit** — Skipping label measurement for percent stacks is correct: `maxWidth` is only read by `calculateXAxisPadding` in `x-scale.ts`, under `if (!inside)`, and `inside` is forced to `true` for percent stacks; `maxHeight` isn't read anywhere. Without context, though, it looks like a bug. Could we add a short comment explaining why? Also, `percentage: undefined` here is redundant and could be dropped.

src/core/shapes/bar-y/prepare-data.ts
```diff
@@ -94,15 +94,20 @@ export async function prepareBarYData(args: {
                 : measureValues;
 
             let ratio = 1;
+            let percentTotal = 0;
             if (series.some((s) => s.stacking === 'percent')) {
+                percentTotal = sortedData.reduce((acc, item) => {
```
korvin89: 🤖 **AI generated**

🔵 **nit** — `percentTotal` keeps only finite values `> 0`, but `sum` a few lines below (`:104`) uses a truthy check (`if (item.data.x)`), so negative values would still count there. Since negatives are now rejected, the impact is small, but it would be nice if both used the same filter.

korvin89 (approved):

PR #673, opened 2026-08-27
feat: add area range series

nikita-jpg:
## What

- Adds an area-range series with `y0` and `y1` values for each point.
- Supports solid and gradient line/fill colors, opacity, null handling, data labels, legends, range sliders, zoom, and multiple Y axes.
- Excludes incomplete points from the axis domain and orders paths by the effective category positions.
- Formats both boundaries independently in tooltips and data labels.
- Uses the range width (`y1 - y0`) for built-in tooltip sorting and totals.
- Moves axis-domain extraction and zoom capabilities into the series plugin API so each series declares its own behavior.
- Adds validation, ChartConfig schema support, Storybook and documentation examples, and focused unit and visual coverage.

## Why

The goal is to add the equivalent of the Highcharts area range chart to this library, implemented natively through its series plugin architecture. Highcharts `low` and `high` values map to `y0` and `y1`. Area range charts are useful for visualizing intervals such as uncertainty bands, confidence ranges, and min/max observations while preserving both boundaries as first-class data.

## Reference

- Highcharts area range example: https://www.highcharts.com/demo/highcharts/arearange

## Testing

- npm run typecheck
- npm run lint:styles
- npm run lint:other
- ESLint on all changed TypeScript and TSX files
- npm test -- --runInBand
- npm run test:chart-config
- npm run build
- npm run docs:build-examples
- Playwright hover and click component test in Chromium
- Visual snapshots in Chromium and WebKit


kuzmadom (commented):
Thanks for the contribution. Implementing Area Range as a separate series plugin fits the current architecture well.

### What is done well

- Series-specific behavior is isolated in its own plugin.
- `getAxisDomainValues` is a useful extension point, and migrating `x-range` to it removes a special case from shared code.
- Range-specific fields were not added to the base series types.
- Validation, gradients, tooltips, events, schema, docs, Storybook, and visual tests are included.

### What needs to be addressed

1. **Public API:** please consider `y0/y1` instead of `low/high`. It is consistent with the existing `x-range` API (`x0/x1`) and represents a generic band between two Y values.
2. **Null values:** a point with one null boundary is not rendered, but its other boundary still affects the Y-domain. Incomplete points should be excluded from the domain.
3. **Range slider and zoom:** Area Range does not preserve adjacent points during X filtering, so the band may not reach the viewport edges. Regular chart zoom is also disabled. These capabilities should be declared by `SeriesPlugin`, not added to shared type-based switches.
4. **Category ordering:** points are not sorted according to the effective category order, which can produce a self-intersecting path.
5. **Formatting:** tooltip custom formatters may be applied to an already formatted range, while numeric data-label formatting does not format both boundaries individually.
6. **Tooltip semantics:** built-in value sorting and totals currently have no defined value for a range.

Please add focused tests for these cases, especially partial nulls, category ordering, range-slider boundaries, and custom formatting.

Overall, the implementation has a solid foundation, but these API and range-specific edge cases should be resolved before merging.

src/plugins/area-range/index.ts
```diff
@@ -0,0 +1,144 @@
...
+} from '~core/validation/helpers';
+
+import {CHART_ERROR_CODE, ChartError} from '../../libs';
+import type {AreaRangeSeries, TooltipDataChunkAreaRange} from '../../types';
+
+import {prepareAreaRangeSeries} from './prepare-area-range-series';
+
+export const areaRangePlugin: SeriesPlugin<AreaRangeSeries> = {
+    type: 'area-range',
+    zoom: {types: ['x', 'xy', 'y'], defaultType: 'x', preserveAdjacentPoints: true},
```
korvin89: 🤖 **AI generated**

🔴 **blocker**

This plugin declares `zoom.types: ['x', 'xy', 'y']`, but the Y filter in `src/core/zoom/zoom.ts:154` reads `point.y`. Area-range points only have `y0`/`y1`, so every point fails the check: `getZoomedSeriesData` with `zoomState.y` returns `[]` for both `preparedSeries` and `preparedShapesSeries`.

What users would see:
- **Area-range alone:** a Y or XY brush does nothing.
- **Next to a line series:** the zoom is applied and the band disappears.
- **Next to a scatter series:** `getZoomType([scatter, areaRange])` defaults to `xy`, so any zoom removes the band. Scatter points with a confidence band are probably the most common use of this type.

Could we handle `y0`/`y1` in that filter and keep a point whose `[y0, y1]` overlaps `[yMin, yMax]`, like the existing `x0`/`x1` branch? If that's out of scope for this PR, declaring `types: ['x']` for now would also work. Either way, a Y/XY zoom test would be great.

src/plugins/area-range/index.ts
```diff
@@ -0,0 +1,144 @@
...
+
+            return y0 === null || y1 === null ? null : y1 - y0;
+        },
+        rows: [
+            {
+                id: 'default',
+                cells: [
+                    {
+                        id: 'color',
+                        source: ({item}) =>
```
korvin89: 🤖 **AI generated**

🟠 **should-fix**

The color cell's `source` returns the swatch markup from `getTooltipColorSymbol(...)`. Area and line used this pattern before #687. On main they now use `source: 'color'` plus a custom `format` that renders the swatch. #687 also documents that `ChartTooltipRowRendererArgs.color` is always a plain CSS color. After a rebase, a custom `rowRenderer` would receive SVG markup for area-range only.

Could we copy the color cell from the area plugin on main and add `area-range` to the `test.each` in `DefaultTooltipContent.test.tsx`? The branch also conflicts with main in that test file right now, so it needs a rebase anyway.

src/core/shapes/area-range/prepare-data.ts
```diff
@@ -0,0 +1,138 @@
...
+            markers: [],
+            opacity: item.opacity,
+            points,
+            series: item,
+            svgLabels: [],
+            width: item.lineWidth,
+        };
+
+        if (!isRangeSlider && shouldPrepareSeriesDataLabels(item)) {
+            const labelPoints = points.map((point) => ({
```
korvin89: 🤖 **AI generated**

🟠 **should-fix**

When a point has `label`, `dataLabels.format` is ignored. Here's why:
- The formatted text goes into `data.y`, but the spread keeps the raw `data.label`.
- `preparePointDataLabels` reads `label ?? y` with `format: undefined`.

For example, `label: 1234.5678` with `{type: 'number', precision: 1}` renders `"1234.5678"`, while line and area apply the format to `label`. This also means the `label` branch in `formatAreaRangeDataLabel` never has any effect.

Maybe override `label` in `labelPoints` instead of `y`, e.g. `label: formatAreaRangeDataLabel({...})`?

src/core/shapes/area-range/prepare-data.ts
```diff
@@ -0,0 +1,138 @@
...
+import type {ChartScale} from '../../scales/types';
+import type {PreparedAreaRangeSeries} from '../../series/types';
+import {getXValue, getYValue} from '../../shapes/utils';
+import {preparePointDataLabels, shouldPrepareSeriesDataLabels} from '../../utils';
+import {createGradientColorResolver, getGradientBBox} from '../../utils/gradient';
+
+import {formatAreaRangeDataLabel} from './format';
+import type {AreaRangePointData, PreparedAreaRangeData} from './types';
+
+function getRangeBBox(points: AreaRangePointData[]) {
```
korvin89: 🤖 **AI generated**

🟠 **should-fix**

`getRangeBBox` adds each boundary as a separate point. In `skip` mode, an incomplete point like `{y0: 90, y1: null}` still expands the gradient bbox even though it isn't drawn; area's `computeAreaBBox` skips such points. With a gradient, this shifts the colors of the rendered band. The same helper is also duplicated in `renderer.ts:24`.

Could we skip points where either boundary is `null`, and share a single helper between the two files?

korvin89 (commented):
🤖 **AI generated**

**Severity legend**

src/core/series/seriesRegistry.ts
```diff
@@ -1,14 +1,15 @@
 import {ChartError} from '../../libs';
+import type {ChartSeries, MeaningfulAny} from '../../types';
 
 import type {SeriesPlugin} from './plugin';
 
-const registry = new Map<string, SeriesPlugin>();
+const registry = new Map<string, SeriesPlugin<MeaningfulAny>>();
```
korvin89: 🤖 **AI generated**

🟠 **should-fix**

The registry is widened to `SeriesPlugin<MeaningfulAny>`, which makes every `getSeriesPlugin()` result effectively untyped. As far as I can tell, the only reason is that `SeriesAxisDomainValues.x`/`y` in `src/core/series/plugin.ts:25` are function-typed properties, which TypeScript checks contravariantly. With main's registry typing, `registerSeriesPlugin(linePlugin)` fails to compile.

I'd suggest declaring them as methods, like the existing `getColorValue?`:

```ts
export interface SeriesAxisDomainValues<T extends ChartSeries> {
    x?(data: T['data'][number]): AxisDomainValue | AxisDomainValue[];
    y?(data: T['data'][number]): AxisDomainValue | AxisDomainValue[];
}
```

Then the registry can stay `SeriesPlugin`. The only other change is a cast for `d` in `getDomainDataXBySeries`/`getDomainDataYBySeries`. With this change, `tsc --noEmit` passes.

src/core/types/chart/area-range.ts
```diff
@@ -0,0 +1,30 @@
+import type {SERIES_TYPE} from '../../constants';
+import type {MeaningfulAny} from '../misc';
+
+import type {BaseSeries, BaseSeriesData, BaseSeriesLegend} from './base';
+import type {SeriesColor} from './gradient';
+import type {RectLegendSymbolOptions} from './legend';
+import type {ChartSeriesRangeSliderOptions} from './series';
+
+export interface AreaRangeSeriesData<T = MeaningfulAny> extends BaseSeriesData<T> {
```
korvin89, line 9: 🤖 **AI generated**

🟠 **should-fix**

The new public types have no JSDoc, while `AreaSeries` documents every field. The guide links to the generated `AreaRangeSeries` API page, which currently has no descriptions. It would be great to document:
- what `y0` and `y1` mean, and the `y0 <= y1` constraint;
- the `nullMode` values and their default (`'skip'`);
- `opacity` (default `0.75`) and `lineWidth` (default `1`);
- `fillColor` and `rangeSlider`.

Related gaps:
- `ChartSeriesOptions['area-range']` has no docs.
- The supported and default zoom types listed in `ChartZoom.type` don't mention area-range.
- The `tooltip.sorting.key` docs don't say that area-range uses the range width.
- `TooltipRowCellItem.formatValue` has no docs.

src/plugins/area-range/__tests__/area-range.test.ts
```diff
@@ -0,0 +1,78 @@
...
+        name: 'Area range',
+        nullMode: 'skip',
+        opacity: 0.75,
+        type: 'area-range',
+        visible: true,
+        yAxis: 0,
+    } as PreparedAreaRangeSeries;
+}
+
+describe('area-range plugin', () => {
```
korvin89, line 53: 🤖 **AI generated**

🟠 **should-fix**

The previous review asked for tests around partial nulls. Right now they're covered only at the Y-domain level. Could we add tests for:
- Y/XY zoom on area-range (this would have caught the zoom issue);
- rendering with `nullMode: 'skip'` (a gap) vs `'connect'`;
- tooltip lookup skipping incomplete points;
- the x-range move to `getAxisDomainValues.x`;
- `preserveAdjacentPoints` on line and area, now that it's a plugin flag.

src/core/types/chart/tooltip.ts
```diff
@@ -196,6 +215,7 @@ export interface TooltipRowCellItem {
      * `cell.format` → `series.tooltip.valueFormat` → `tooltip.valueFormat`.
      */
     format?: ValueFormat;
+    formatValue?: (args: TooltipRowCellFormatValueArgs) => string;
```
korvin89: 🤖 **AI generated**

🟠 **should-fix**

`formatValue` is added to the public (experimental) `TooltipRowCellItem` for a single plugin. It has no docs and no defined precedence relative to `format`, and once it's released it will be hard to take back. Also, `tooltip.getValue` (`src/plugins/area-range/index.ts:104`) and the value cell's `source` both compute `y1 - y0`.

Could we keep `formatValue` on an internal cell type used only by plugin `TooltipRowDef`s? If it's meant to be public, it would need JSDoc and a documented precedence. Sharing one helper for the width would also be nice.

src/components/Tooltip/DefaultTooltipContent/index.tsx
```diff
@@ -173,10 +174,16 @@ export const DefaultTooltipContent = ({
...
-                            }),
+                            formattedValue: valueCell?.formatValue
+                                ? valueCell.formatValue({
+                                      item: seriesItem,
+                                      value,
+                                      format: valueCell.format ?? rowValueFormat,
+                                  })
+                                : getFormattedValue({
+                                      value,
+                                      format: valueCell?.format ?? rowValueFormat,
```
korvin89: 🤖 **AI generated**

❓ **question**

Two things about the tooltip value for area-range:

1. A custom `rowRenderer` gets the range width (`y1 - y0`) as `value`, but the range string as `formattedValue`. The "Sum" total also adds up widths. Is that intended? If so, it might be worth documenting.
2. The `rowRenderer` path now formats with `valueCell?.format ?? rowValueFormat` instead of `rowValueFormat`. This affects user `tooltip.rows` that set both `renderer` and a value-cell `format`. It also seems to contradict the `ChartTooltipRow` docs: "If provided, `cells` is ignored for that row." Is this change intentional?

src/core/shapes/area-range/prepare-data.ts
```diff
@@ -0,0 +1,138 @@
...
+                y0: absoluteY0,
+                y1: absoluteY1,
+                y,
+                color: data.color,
+                data,
+                series: item,
+            });
+        }
+
+        points.sort((a, b) => a.x - b.x);
```
korvin89, line 74: 🤖 **AI generated**

🟠 **should-fix**

Area and line mark points outside the visible Y range with `markHiddenPointsOutOfYRange`, and their `getTooltipData` skips those points. Area-range does neither. So when `yAxis.min`/`max` puts a band entirely outside the plot, hovering there still shows a tooltip for a point that isn't visible. The tooltip shouldn't appear in that case.

Could we mark points whose whole `[y0, y1]` interval lies outside the plot range as hidden, keeping neighbors as anchors like area does? `getTooltipData` could then skip them.

src/core/series/plugin.ts
```diff
@@ -96,6 +116,8 @@ export interface SeriesPlugin<T extends ChartSeries = ChartSeries> {
      * Omit for types that do not support a continuous color scale (e.g. treemap, sankey, radar).
      */
     getColorValue?(data: T['data'][number]): number | string | null | undefined;
+    getAxisDomainValues?: SeriesAxisDomainValues<T>;
```
korvin89, line 138: 🤖 **AI generated**

🟠 **should-fix**

The new `SeriesPlugin` members `getAxisDomainValues`, `zoom` and `tooltip.getValue` have no JSDoc, while the rest of the interface is documented in detail. `preserveAdjacentPoints` especially needs a comment, because it isn't obvious that the flag is ignored on a category x-axis. Also, `zoom` currently sits in the "Data preparation" section.

src/core/shapes/area-range/format.ts
```diff
@@ -0,0 +1,24 @@
+import type {AreaRangeSeriesData, ValueFormat} from '../../../types';
+import {getFormattedValue} from '../../utils/format';
+
+export function formatAreaRange(args: {y0: number; y1: number; format?: ValueFormat}) {
+    const {y0, y1, format} = args;
+    const formattedY0 = getFormattedValue({value: y0, format});
+    const formattedY1 = getFormattedValue({value: y1, format});
+
+    return `${formattedY0} – ${formattedY1}`;
```
korvin89: 🤖 **AI generated**

🟠 **should-fix**

The range separator here is an en dash (`" – "`), while x-range already uses an em dash (`" — "`, `src/plugins/x-range/index.ts:79`). Could we reuse the existing x-range separator so ranges look the same across series types? A few tests match on `–` and would need updating:
- `src/__tests__/area-range.visual.test.tsx`
- `AreaRangeEventsTestStory`
- `area-range.test.ts`
- `DefaultTooltipContent.test.tsx`

src/plugins/area-range/index.ts
nikita-jpg: Fixed in 8e8ef1e. The area-range plugin now provides an interval-aware Y filter: a complete point is kept when [y0, y1] overlaps the zoom range, including when it contains the entire viewport. Added Y/XY filtering tests and browser coverage for area-range alone, with line, and with scatter.

src/plugins/area-range/index.ts
nikita-jpg: Updated the color cell to source: 'color' with swatch rendering in format, matching the area plugin on main. Added area-range to the raw-color rowRenderer test matrix. The branch is rebased onto main and the test-file conflict is resolved.

src/core/shapes/area-range/prepare-data.ts
nikita-jpg: Fixed through an internal getLabelText callback in label preparation. Explicit labels now use dataLabels.format exactly once, while the original point and its label remain unchanged. Added regression tests for numeric labels in both SVG and HTML, plus a custom formatter call-count test.

src/core/shapes/area-range/prepare-data.ts
nikita-jpg: Fixed. Preparation and rendering now share one getRangeBBox helper, which excludes a point if either boundary is null. Added a regression test verifying that incomplete points change neither the bounding box nor the colors of complete points.

src/core/series/seriesRegistry.ts
nikita-jpg: Applied the method signatures for SeriesAxisDomainValues.x/y and restored the registry and getSeriesPlugin return type to SeriesPlugin. The domain helpers now cast the input at the dispatch boundary. Typecheck passes.

src/core/types/chart/area-range.ts
nikita-jpg: Added API documentation for the boundaries and their constraint, nullMode and its default, opacity, lineWidth, fillColor, rangeSlider, and the area-range series options. Updated the zoom and tooltip sorting/totals documentation as well. formatValue was removed from the public cell type and kept internal. The declaration/JSON Schema checks and docs preview pass.

src/plugins/area-range/__tests__/area-range.test.ts
nikita-jpg: Added coverage for all five cases: Y/XY zoom; rendered gaps versus connections for partial nulls; tooltip lookup excluding incomplete points; x-range domain extraction through the shared pipeline; and adjacent-point preservation for line and area on continuous X axes, with category-axis behavior checked separately. Y/XY zoom also has browser tests for area-range alone and mixed with line/scatter.

src/core/types/chart/tooltip.ts
nikita-jpg: Moved formatValue to the internal PluginTooltipRowCell type used by TooltipRowDef; it is no longer part of public TooltipRowCellItem. Also extracted getAreaRangeWidth and reused it for tooltip.getValue, the value cell source, and the color-scale value.

src/components/Tooltip/DefaultTooltipContent/index.tsx
nikita-jpg: 1. Yes, the intended scalar value is the range width (y1 - y0), while the default display formats both boundaries. This is now documented for rowRenderer, sorting, and totals; sum adds widths, not interval unions. Added tests for the renderer arguments and totals.

2. The formatting-precedence change was unintended and has been reverted. Custom row renderers use rowValueFormat, without applying the value cell's format. Added a regression test for a row that supplies both renderer and a cell format.

src/core/shapes/area-range/prepare-data.ts
nikita-jpg: Fixed. Complete intervals entirely outside the visible Y range are excluded from tooltip lookup. Shape visibility is tracked separately so neighboring points can remain as path anchors. Added tests for partially visible intervals, off-screen neighbors, and explicit Y limits with a clamped scale.

src/core/series/plugin.ts
nikita-jpg: Documented getAxisDomainValues, zoom, tooltip.getValue, and the zoom options, including preserveAdjacentPoints being ignored for category X axes. Moved zoom alongside the plugin metadata instead of leaving it in data preparation.

src/core/shapes/area-range/format.ts
nikita-jpg: Changed the separator to the same em dash (' — ') used by x-range and updated the formatter assertions and click-event example. The hover/click browser test passes with the new separator.

src/core/shapes/area-range/prepare-data.ts
```diff
@@ -0,0 +1,127 @@
...
+                    }
+                });
+            }
+        }
+
+        const prepared: PreparedAreaRangeData = {
+            active: true,
+            annotations: [],
+            color: item.color,
+            getHoverMarkers: () => [],
```
korvin89: 🤖 **AI generated**

🟠 **should-fix**

Area-range should have marker parity with area. Right now it has no markers at all:
- `getHoverMarkers` returns `[]`;
- `markers` is always empty;
- neither `AreaRangeSeries` nor `ChartSeriesOptions['area-range']` has a `marker` option.

By default, hovering a band only brightens it.

For comparison, area:
- hides normal markers by default (`DEFAULT_MARKER.enabled: false`);
- shows a hover marker by default (`states.hover.marker`: enabled, white 1px border, halo), rendered through `buildHoverMarkerGetter`;
- lets users configure markers via `series.marker`, per-point `data[].marker`, `seriesOptions.area.marker`, and `seriesOptions.area.states.hover.marker`.

Could we bring the same behavior to area-range, with markers on both boundaries (`y0` and `y1`) of the hovered point? Concretely:
- `marker` on `AreaRangeSeries` and per-point `marker` on `AreaRangeSeriesData`, as in area;
- `marker` and `states.hover.marker` (including `halo`) in `ChartSeriesOptions['area-range']`;
- normal markers (when enabled) and hover markers on both boundaries, reusing `buildHoverMarkerGetter` and `getMarkerFill` where possible;
- tests for the default hover markers and for the options above.

korvin89: 🤖 **AI generated**

🟠 **should-fix**

Thanks for the quick turnaround! First, a correction on my side: my previous comment said area shows a halo on hover, and it doesn't. Area prepares `states.hover.marker.halo`, but `renderHoverMarkers` only draws the symbol, so neither area nor line renders a halo. Sorry for the confusion.

Because of that, area-range now goes beyond area, and the shared marker code was changed to support it. Could we keep this PR to strict parity with area? The extras could go into a separate PR that covers line, area and area-range together.

**Keep:**
- **Types:** `AreaRangeSeries.marker`, `AreaRangeSeriesData.marker`, and `marker` / `states.hover.marker` in `ChartSeriesOptions['area-range']`. These are the same types as area, `halo` included.
- **Option preparation** in `prepare-area-range-series.ts`, following area's `prepareMarker`:
  - normal markers are disabled by default;
  - the hover marker is enabled, with the normal radius and a white 1px border;
  - `states.hover.marker` is merged on top.
- **Normal markers** on both boundaries when `marker.enabled` or a point's `marker.states.normal.enabled` is set. The fill goes through `getMarkerFill` in this order:
  - point `marker.color`;
  - point `color`;
  - the gradient sampled at the boundary;
  - the series color.
- **Hover markers** on both boundaries through `buildHoverMarkerGetter`, only when normal markers are disabled, as in area.
- **Edge cases:** one marker for coincident boundaries; no markers for incomplete points or for boundaries outside the plot.
- **Tests, docs and schema changes** for all of the above.

**Remove:**
- **Halo rendering:** `MarkerItem.halo`, the halo branch in `setMarkerStyles`, and `halo` in `getHoverStyle` (`markers.ts`).
- **Hover restyling of permanent markers:** `MarkerItem.hover`, the `hover-shape.${namespace}-normal` subscription in `MarkerLayer`, and the `dispatcher`/`namespace` props passed from `SeriesShapes`. With permanent markers, hover should change nothing, as in area.
- **Extra color sources:** the `states.hover.marker.color` override for hover markers, and the series-level `marker.color` (`series.marker.color`, `series.options['area-range'].marker.color`) as a fill source. Area ignores both; its normal state omits `color`.
- **Shared-file changes:** after the removals above, `getHoverMarkers` can return the `buildHoverMarkerGetter` results as they are, so `getHoverStyle` is no longer needed. `src/core/shapes/marker.ts`, `src/core/shapes/types.ts`, `src/hooks/useShapes/MarkerLayer.tsx` and `src/hooks/useShapes/SeriesShapes.tsx` go back to main.
- **Docs and tests** that mention halos or hover/series marker colors:
  - the `halo` options in the guide and in the docs example;
  - the halo and permanent-marker hover assertions in `area-range.visual.test.tsx`;
  - the same assertions in `area-range.test.ts`.

korvin89 (approved):
🤖 **AI generated**

Thanks for working through all the feedback — marker behavior now matches area, and the shared marker code is back to main.

PR #681, opened 2026-08-30
fix(area): preserve raw tooltip color for custom row renderer

kuzmadom:


kuzmadom:
duplicate https://github.com/gravity-ui/charts/pull/687

PR #683, opened 2026-09-01
feat(scatter): add point clustering

nikita-jpg:
## Summary

- Adds optional pixel-distance clustering for scatter series.
- Exposes the cluster size and original points through tooltip and chart click data.
- Supports configurable cluster distance, minimum cluster size, marker appearance, and count labels.
- Keeps clustering disabled by default.
- Leaves range-slider rendering unclustered.

## Motivation

Dense scatter plots can contain many overlapping markers, making individual regions difficult to read and interact with. Grouping nearby points into a single marker keeps the chart legible while preserving access to the original points for tooltips and custom interactions.

## API

Clustering is enabled through `series.cluster` and supports:

- `enabled`
- `distance`
- `minimumSize`
- marker color and radius
- cluster-count label format and style

Cluster interaction data includes:

- `clusterSize`
- `clusteredData`
- cluster coordinates and radius

## Testing

- `npm run typecheck`
- `npm test -- --runInBand`
- `npm run build`
- Chromium component test covering rendering, hover tooltip, click data, and visual snapshot


korvin89:
🤖 **AI generated**

Thanks for taking this on — clustering is a genuinely useful addition, and the placement and most of the naming here are already close to the prior art. Before going through the implementation, I'd like to agree on the public API: a few things below would change the shape of the config and the data contract, and it seems better to settle that first than to review the details twice.

I've held back the code-level comments for now. There are some separate correctness concerns I'll write up once the interface is agreed — including one about the clustering algorithm itself, which is relevant to point 1.

## What we'd change

**1. `distance` sits at the top level with no algorithm selector.** It's an algorithm-specific parameter promoted to the top of the namespace. In Highcharts it lives at `cluster.layoutAlgorithm.distance` and is documented as applying to `kmeans` only, with `gridSize` as the grid equivalent — nested precisely so the two never end up side by side as mutually exclusive top-level options. If a second algorithm is ever wanted here, `cluster.{distance, gridSize}` at one level is hard to undo. The current implementation is single-linkage, and I have concerns about how it behaves on dense data that I'll send separately — which is part of why naming the algorithm now matters. I'd suggest nesting `distance` under `layoutAlgorithm` with a `type` beside it, even if it ships with one value.

**2. `ScatterClusterMarkerOptions` re-declares part of a type we already have.** It's a two-field copy of `PointMarkerOptions` (`color`, `radius`), and it's currently the only marker options type in the repo that doesn't reuse it — compare `AreaMarkerOptions extends PointMarkerOptions` (`area.ts:53`), `RadarMarkerOptions extends PointMarkerOptions` (`radar.ts:22`), `LineSeries.marker?: PointMarkerOptions` (`line.ts:118`). Beyond the drift risk, it means a cluster marker can never get its own `symbol` or border — and `symbol` is the likely next request, since Highcharts gives clusters a dedicated `'cluster'` symbol so they don't just read as an inflated data point.

**3. `ScatterClusterDataLabelsOptions` re-declares part of `BaseDataLabels`.** I think this type should exist — a count centred inside a marker really is a different thing from a value label above a point — but derived rather than hand-written. Of the omitted fields, `allowOverlap` is the one that actually applies; `html` and `padding` don't apply to a centred count, and a `Pick` states that explicitly instead of leaving it ambiguous. It also gives us somewhere to put the one hook that's genuinely missing — see the aggregation note below.

**4. The cluster payload is flat, prefixed, and carries render values.** `clusterSize` / `clusteredData` sit directly on the point object handed to `chart.events.click` and the tooltip — the same slot as an ordinary `ScatterSeriesData` — so the prefix is currently the only thing distinguishing them. Nesting them under a single `cluster` key removes the stutter, mirrors the option name, and gives a real discriminant: `'cluster' in data` narrows in TypeScript and would replace the structural `'clusteredData' in data` checks. Separately, `radius` and the resolved `color` are render values in what is otherwise a data contract — those feel like they belong on the prepared point, not the public type.

**5. `ScatterClusterData` should extend `ScatterSeriesData` rather than parallel it.** As it stands, widening `TooltipDataChunkScatter['data']` to a union is a compile-time break for consumers reading public `ScatterSeriesData` fields such as `label` or `opacity` in a custom `tooltip.renderer` — even with clustering switched off. Extending keeps the union assignable.

**6. The tooltip reuses the value column for the count.** A cluster's row currently shows `clusterSize` where a point's row shows `y`, through the same `'value'` cell — so a count inherits whatever value formatting the series has. Highcharts deliberately keeps these apart with a separate `tooltip.clusterFormat` rather than overloading the value slot. I think the count wants its own channel in the contract; the good news is that no new public API is needed for it (see the aggregation note).

**7. There's no chart-level default.** Would it make sense to accept `cluster` in `ChartSeriesOptions['scatter']` too, with the series-level value overriding it? The two-level structure already exists here — `prepareScatterSeries` merges `scatter.states.hover` from `seriesOptions`, and `radar`/`line` already expose `marker?: PointMarkerOptions` at that level. Otherwise a chart with eight scatter series repeats the block eight times.

## Proposed shape

```ts
export type ScatterClusterAlgorithm = 'grid' | 'single-linkage';

export interface ScatterClusterLayoutAlgorithmOptions {
    /**
     * Maximum distance in pixels between two points for them to end up in the
     * same cluster. For `grid` it is also the cell size.
     * @default 40
     */
    distance?: number;
    /** The algorithm used to combine nearby points into a cluster. */
    type?: ScatterClusterAlgorithm;
}

export interface ScatterClusterDataLabelsOptions
    extends Pick<BaseDataLabels, 'allowOverlap' | 'enabled' | 'format' | 'style'> {
    /**
     * Value rendered in the cluster label. Receives the cluster point, so any
     * aggregate over `cluster.points` can be returned — a count, a sum, `3 of 12`.
     * `format` is applied to the result.
     * @default the number of clustered points
     */
    source?: (args: {point: ScatterClusterData}) => string | number;
}

export interface ScatterClusterOptions {
    /** Options for the label showing the number of clustered points. */
    dataLabels?: ScatterClusterDataLabelsOptions;
    /**
     * Whether to combine nearby points into clusters.
     * @default false
     */
    enabled?: boolean;
    /** Options for the algorithm used to combine points into a cluster. */
    layoutAlgorithm?: ScatterClusterLayoutAlgorithmOptions;
    /** Options for the cluster marker. Falls back to the series marker. */
    marker?: PointMarkerOptions;
    /**
     * The minimum number of points to be combined into a cluster.
     * Must be greater than or equal to 2.
     * @default 2
     */
    minimumClusterSize?: number;
}

export interface ScatterClusterData<T = MeaningfulAny> extends ScatterSeriesData<T> {
    /** Present only on clustered points. */
    cluster: {
        /** The source points combined into this cluster. */
        points: ScatterSeriesData<T>[];
        /** The number of source points combined into this cluster. */
        size: number;
    };
}
```

## On aggregation

One thing I'd suggest **not** adding: a declarative aggregation config. supercluster and MapLibre need `map`/`reduce` and `clusterProperties` because they never hand you the source points — here `cluster.points` is right there, so aggregation can stay a read-side concern, composed from hooks this library already has:

- **Tooltip** — nothing new required. `TooltipRowCellItem.source` is already public and already takes `({item}) => unknown`, so `source: ({item}) => sum(item.data.cluster.points.map((p) => p.y))` works as soon as `cluster.points` exists; `tooltip.renderer` and `row.renderer` receive the chunk too. On the plugin side, `SeriesPlugin['tooltip']['rows']` already accepts a function of the chunk — the documented precedent is waterfall's conditional "Subtotal" row — so the cluster-specific row from point 6 needs no new public API either.
- **Label** — this is the one gap, and the reason for `source` in the type above. `cluster.dataLabels.format` currently receives only the bare count, and `CustomFormat.formatter` is typed `({value}) => string` library-wide, so widening its arguments isn't a local change. Mirroring the tooltip's `source` + `format` composition keeps it consistent with an idiom that already exists here.
- **Centroid** — leave it as the arithmetic mean and non-configurable, as Highcharts does.

## Architectural question: one labelling mechanism or two?

The count is currently built as a parallel mechanism — its own field on the prepared shape data, its own label type that deliberately isn't `LabelData`, its own construction path, and its own styling path in the renderer — while the cluster itself is an ordinary point that flows through the same `scatterData` and the same `renderMarker` as any other.

Looking at how other libraries split this, the pattern tracks exactly that distinction:

| | mechanism for the count |
|---|---|
| Highcharts — cluster is a `Point` | **the same** dataLabels system, different defaults (`inside: true`, `verticalAlign: 'middle'`, `color: 'contrast'`) |
| Leaflet.markercluster — cluster is an icon | separate (`iconCreateFunction`) |
| Plotly `scattermap` — cluster is a symbol layer | separate |
| amCharts 5 — cluster is a bullet | separate (`clusteredBullet`) |

The libraries with a separate mechanism are the ones where a cluster is a different rendered object. Here it isn't, which puts us in the first row. Reusing the existing label pipeline with cluster-specific defaults would also mean the count participates in the same overlap handling and theming as every other label, rather than needing its own answers to those.

Was the parallel path a deliberate choice, or mostly the shortest route to a working count? Happy either way, but it affects points 3 and 6 above.

## Open questions

1. Do we want `layoutAlgorithm.type` in the public shape from the start, and which algorithm should be the default?
2. What should `x` / `y` mean on a cluster when the axis is `category`, or non-linear such as `logarithmic`? Averaging the underlying values doesn't produce a meaningful result in either case.
3. One labelling mechanism or two?
4. Chart-level `cluster` default — yes or no?
5. Should cluster size be expressible visually — Highcharts `zones`, Plotly's `step` plus array `size`/`color`, Leaflet's 10/100 buckets? Right now a cluster of 3 and a cluster of 500 look identical.
6. Should the tooltip be able to list the clustered points, or stay count-only?

This changes a public API surface, so we'll wait for @kuzmadom to review the shape before you start on the rework.

nikita-jpg:
Thanks for the detailed review. I checked our existing Highcharts usage to clarify what the migration actually needs.

I agree with reusing the marker/data-label types, making `ScatterClusterData` extend `ScatterSeriesData`, nesting the payload as `cluster: {size, points}`, keeping computed marker styling on prepared points, and handling cluster count separately from the regular tooltip value.

My proposal for the open questions:

1. **Algorithm:** our Cycle Time widget currently uses `grid` with `gridSize: 50`. I suggest shipping grid first and making it the default, with options under `layoutAlgorithm`. A cell size should be documented as such, rather than as a maximum pairwise distance. I reproduced the chaining concern: the current implementation merges ten points spaced 30 px apart, spanning 270 px, into one cluster. We also currently use `cluster.allowOverlap: false` to prevent marker collisions; `dataLabels.allowOverlap` would not replace that behavior.
2. **Axes:** we need datetime X and linear Y. Category/logarithmic clustering can be deferred until their coordinate semantics are defined. The source points must retain their original coordinates and `custom` data, including issue IDs.
3. **Labels:** I favor reusing the existing label pipeline with cluster-specific defaults. Our current label is just the count. `source` + `format` makes sense for custom aggregates, but is not required by our migration.
4. **Chart-level defaults:** useful, but not required for our current widget; I suggest a follow-up.
5. **Size-dependent styling:** also a follow-up. We currently use a fixed circular marker with configurable color/radius and a count label.
6. **Tooltip and clicks:** the default tooltip can remain count-only, but our custom tooltip must be able to list and sort the source issues and link to them. On mobile, tapping a cluster opens an issue sheet without automatically zooming. `cluster.points` in the existing tooltip renderer and chart click event should cover these cases; no separate cluster-click API is necessary.

I agree that a declarative aggregation API is unnecessary. I would also include datetime, resize/zoom regrouping, and clicks with tooltips disabled in the regression coverage. I'll wait for agreement on the API before reworking the implementation.

PR #684, opened 2026-09-01
feat(axis): support explicit tick values

nikita-jpg:
## Summary

- Adds `ticks.values` for explicitly specifying X and Y axis tick positions.
- Supports numeric, datetime, and category axes.
- Preserves the specified values without automatic tick thinning.
- Filters out values outside the current axis domain.
- Allows an empty array to suppress all ticks.

## Motivation

Some charts require stable, domain-specific tick positions instead of automatically selected values. Explicit tick values make axis rendering deterministic and provide functionality similar to `tickPositions` in other charting libraries.

## API

```ts
const data = {
    xAxis: {
        ticks: {
            values: [0, 4, 10],
        },
    },
    yAxis: [
        {
            ticks: {
                values: [0, 40, 100],
            },
        },
    ],
};


korvin89:
🤖 **AI generated**

Thanks for this — explicit tick positions are a genuinely useful addition, and the API shape is a good fit. `ticks.values` matches `axis.values` in Vega-Lite and reads as `tickValues` (D3) once the namespace is included, which is nicer than Highcharts' `tickPositions`, where "positions" misleadingly suggests pixels. Extracting the inline type into an exported `ChartAxisTicks` is a clean, backwards-compatible refactor too.

Before going further I'd like to agree on the semantics, because there is one structural issue that shows up as several separate symptoms. Rather than leaving these as scattered review comments, here is what I think the feature should do.

### The underlying issue

`getTickValues` (and `getXAxisTickValues`) has two distinct jobs in this codebase:

1. produce the ticks that get rendered;
2. serve as the input from which the Y domain is derived — `getDomainMinAlignedToStartTick` / `getDomainMaxAlignedToEndTick` (`src/core/scales/y-scale.ts:156,205`) and `primaryTicksCount` for the secondary-axis sync (`src/hooks/useAxisScales/index.ts:83`).

Explicit values are only meaningful for job 1. Because they currently flow into job 2 as well, the domain code ends up aligning the extent to an arbitrary step taken from the user's array. Three reproducible consequences:

- `order: 'reverse'` + `startOnTick: true` + `values: [0, 40, 100]` on `min: 0 / max: 100` produces an **inverted domain `[160, 100]`** (baseline without `values`: `[0, 100]`). The result is sorted descending by *position*, so on a reversed range `tickValues[0]` is the largest value, `step` comes out negative, and `dNewMin = tickValues[0].value - step` overshoots past the max.
- `startOnTick` + `endOnTick` with `values: [20, 40]` on `min: 0 / max: 100` stretches the domain to `[0, 120]`. Since ticks are pinned, nothing lands on 120 — so `endOnTick` is not actually satisfied and the plot just grows by an arbitrary amount. The same happens with `values: [0, 40, 100]` when `max` is left unset: `maxPadding` puts `dMax` at ~105, `step` is 40, the domain becomes 120, and there is ~66px of empty space above the top tick.
- A primary Y axis with `values: [0, 40, 100]` paired with a secondary axis (both `startOnTick` + `endOnTick`) sets `primaryTicksCount` to 3, which blows the secondary domain up from `[-10, 90]` to `[-100, 100]` — 11 ticks aligning with none of the primary's 3 gridlines, and its data (0..73) squeezed into the top half.

Worth noting that the current tests pass only because the story and the visual test both pin `min: 0, max: 100` with `values` that include both endpoints. That is the one configuration where the outermost tick lands exactly on the range edge, so the alignment code short-circuits.

### Proposed semantics

**`ticks.values` selects which ticks are drawn. It never changes the axis extent.**

The domain keeps being derived from the data, `min` / `max`, `maxPadding` and `startOnTick` / `endOnTick`, exactly as today, using the automatic tick generator. Explicit values falling outside the resulting domain are dropped.

Two reasons I'd like to settle on this rather than the Highcharts behaviour, where `startOnTick` / `endOnTick` snap the extent onto the outermost explicit tick:

- The X axis already works this way — `createXScale` derives `startOnTick` / `endOnTick` via `nice()` and never looks at tick values (`src/core/scales/x-scale.ts:178`). So this is about making the Y axis internally consistent with the X axis, not about importing an outside convention.
- It makes the feature strictly additive: adding `ticks.values` to an existing config can change which ticks appear, but can never move the domain. That is a property we can assert in a test.

### How the options compose

The dividing line: `min` / `max` / `startOnTick` / `endOnTick` shape the domain, while `ticks.interval` / `ticks.pixelInterval` / `ticks.values` shape the ticks. `ticks.values` replaces the tick generator's *output*; `interval` and `pixelInterval` remain *hints* to that generator. The domain never sees `values` at all.

- **`min` / `max`** — unchanged. They decide which of your `values` survive clipping. Setting them explicitly also disables `maxPadding` on that side, which is why the story in this PR happens to put ticks exactly on the axis edges.
- **`startOnTick` / `endOnTick`** — unchanged, and deliberately computed from *automatic* ticks, as if `values` were not set. The visible consequence is that the edge of the axis may not carry a tick. That is expected, and should be documented rather than worked around.
- **`ticks.interval` / `ticks.pixelInterval`** — no effect on which ticks are drawn, since `values` wins, but they still influence the domain through `startOnTick` / `endOnTick`, because the domain is derived from automatic ticks. This reads odd in isolation, but it is exactly what "the extent is what it would have been without `values`" means, and I would rather keep that invariant than special-case it. Documenting the precedence is enough; there is no need to reject the combination in `validateAxes`.

Two more things worth adding to the JSDoc. First, `interval` and `pixelInterval` are *pixel* hints — `interval` is resolved against the axis size via `calculateNumericProperty` — whereas `values` are in data units. ECharts' `interval` and Plotly's `dtick` are data-space, so this is an easy trap, and the two options now sit side by side in the same object. Second, the recipe for "make the axis span exactly my tick range" differs per axis: on a linear or logarithmic Y axis, `min` / `max` is enough, since `startOnTick` / `endOnTick` default to `false` there; on the X axis they default to `true`, and `nice()` can widen the domain past an explicit `min` (`Math.min(nicedDomain[0], domainWithOffset[0])`, `src/core/scales/x-scale.ts:184`), so `startOnTick: false` / `endOnTick: false` are needed as well.

### What that means in code

The main change is to separate the two jobs, rather than adding `if (axis.ticks.values)` guards in three places. Concretely: keep the current body of `getTickValues` / `getXAxisTickValues` available as an "automatic ticks" function, have the rendering path return explicit values when present, and point the three domain-derivation call sites (`y-scale.ts:156`, `y-scale.ts:205`, `useAxisScales/index.ts:83`) at the automatic one. That resolves all three symptoms above and preserves the additivity property.

One consequence worth documenting: on a chart with a synced secondary Y axis, the primary's explicit ticks will not line up with the secondary's labels. That is inherent — explicit positions are not evenly spaced, so no choice of secondary domain can make them coincide — and documenting it seems better than special-casing it.

Alongside that, a few things I'd like resolved in the same PR.

**Sort by value, not by position.** The automatic path returns ticks in domain order — ascending value for continuous scales, `scale.domain()` order for band scales. The two adapters currently sort by pixel position, in opposite directions (`x-axis.ts:66` ascending, `AxisY/utils.ts:26` descending), which is what introduces the reversed-axis bug above. Sorting by value in both restores parity with the automatic path in every orientation and removes the asymmetry between the call sites.

**Filter in domain space, not pixel space.** `getExplicitAxisTickValues` tests the scaled position against `scale.range()`, which is not equivalent to "outside the axis domain". On a degenerate domain such as `domain([5, 5])` — a case this codebase handles deliberately, see `checkIsPointDomain` and the `yMin === yMax ? 0 : …` padding branch — `scale(0) === scale(5) === scale(1000)`, so `values: [0, 5, 1000]` renders three ticks stacked at the midpoint, labelled 0, 5 and 1000. Comparing against `scale.domain()` instead, and keeping the `Number.isFinite` check (which is what correctly drops non-positive values on a log scale), would match the documented intent. Separately, the explicit branch is inserted above the existing `if (!axisWidth) return []` / `if (!height) return []` guards, so a zero-sized range now yields every value at position 0 instead of an empty array — an early `if (rangeMax === rangeMin) return []` inside the helper covers both call sites.

**Category indices need a single index space.** Right now there are two. `min` / `max` index the `categories` array as authored — they are consumed by `getAxisCategories`, which applies `order` and then slices. `values` are resolved much later against `scale.domain()`, which is that already-sliced array, further filtered by `filterCategoriesByVisibleSeries` (`src/core/scales/utils.ts:111`). So `categories: ['A', 'B', 'C']` with `min: 1` makes `values: [0]` mean `'B'`, not `'A'`; and with data present only at A and C, the domain is `['A', 'C']`, so `values: [2]` silently renders nothing while `values: [1]` lands on C. The indices also shift whenever a series is toggled in the legend.

I'd like `values` to index the same array `min` / `max` do: the categories as authored. Slicing and visibility filtering then simply *remove* ticks rather than shifting the remaining ones. The cleanest place for that is `getNormalizedXAxis` / `getNormalizedYAxis` (`src/components/ChartInner/utils/normalized-original-data.ts`), which already operates on the original array via `getAxisCategories` — converting the indices to category names there means everything downstream only ever deals with names. That also leaves the door open for accepting `string[]` publicly later, as ECharts, Plotly, Vega-Lite and Recharts all do, since the internal representation would already be names.

**De-duplicate the input.** `values: [50, 10, 50, 10]` currently renders four ticks in two overlapping pairs, and makes `getMinSpaceBetween(...)` return 0, so `tickStep` is 0 and `labelMaxWidth` / `labelMaxHeight` in both prepare-axis-data files collapse to a negative number. ECharts and Vega both de-duplicate and sort explicit values; a `new Set(...)` would do it here.

**Documentation.** `values` is the only field in `ChartAxisTicks` without a JSDoc comment, and the generated API docs read straight from these comments. The `min` / `max` docs a few lines up (`src/core/types/chart/axis.ts:186-196`) are a good template — they already break the value type down per axis scale, and `values` needs the same: a plain value for linear and logarithmic axes, a timestamp for datetime, a category index for category. Worth adding alongside that: ticks are not thinned, values outside the domain are dropped, `[]` hides every tick, and `values` takes precedence over `interval` / `pixelInterval` when both are set.

**Tests.** The claims in the description that currently have no coverage: `values: []` and datetime axes. Also worth pinning down, though not claimed: logarithmic axes, `startOnTick` / `endOnTick`, `order: 'reverse'`, the secondary-Y path, and duplicate or out-of-range input. Most of this is pure logic and would be cheaper and more precise as a unit test on `getExplicitAxisTickValues` next to the existing suites in `src/core/utils/__tests__/` — a Playwright screenshot is a heavy instrument for asserting that NaN is dropped. I'd also suggest moving the Y-axis assertions in the new test out of `x-axis.visual.test.tsx`, since `y-axis.visual.test.tsx` already exists.

Smaller things, entirely at your discretion:

- `getExplicitAxisTickValues` re-implements helpers that live a few lines below it in the same file — `getXTickPosition` / `center` / `number` already handle "band centre plus half the bandwidth" (and additionally honour `scale.round()`) and "coerce through a numeric scale". Reusing them would keep band positioning in one place.
- `const scaleValue = scale as (value: number | Date) => number` casts the `ChartScale` union to a bare call signature; `number(scale)` exists for exactly this. Note that `scaleLog(NaN)` and `scaleUtc(new Date(NaN))` return `undefined` rather than `NaN`, so it is the `Number(...)` wrapper that makes the `Number.isFinite` check work — worth a comment if the cast stays.
- `axis.type === 'datetime' ? new Date(value) : value` is a no-op, since `scaleUtc` accepts a raw timestamp. If it is kept for clarity, keying it off the existing `isTimeScale(scale)` guard would match how the rest of the file discriminates.
- In the band branch, `position >= rangeMin && position <= rangeMax` can never fail, since a band centre is always inside the range; the real filter there is the `undefined` check.
- Field ordering: `ChartAxisTicks` is declared `pixelInterval` (deprecated) / `interval` / `values`, and the two prepared-axis literals put `values` before `pixelInterval` while `PreparedBaseAxis` (`src/core/axes/types.ts:103-104`) declares them the other way round. Alphabetical — `interval`, `pixelInterval`, `values` — throughout would be consistent.

@kuzmadom, could you validate this before the rework starts? The extent question is the one that matters most: whether `ticks.values` should be purely presentational, as proposed here, or should move `min` / `max` onto the outermost explicit tick the way Highcharts does. Everything else follows from that choice.

kuzmadom:
> @kuzmadom, could you validate this before the rework starts? The extent question is the one that matters most: whether `ticks.values` should be purely presentational, as proposed here, or should move `min` / `max` onto the outermost explicit tick the way Highcharts does. Everything else follows from that choice.

I agree with the proposed presentational semantics: `ticks.values` should not affect the axis domain. The current API shape looks sufficient to me.

  Two things worth accounting for during the rework: label layout currently assumes uniform tick spacing, and category `min`/`max` indices are applied after ordering, so the proposed category-index semantics need clarification.

nikita-jpg:
Thanks for the detailed review and for confirming the semantics.

I’ve reworked the implementation:

- `ticks.values` now affects rendering only; domain alignment and secondary-axis synchronization use automatic ticks.
- Explicit values are deduplicated, sorted consistently, and filtered in domain space.
- Category indices are resolved after applying `order`, but before `min` / `max` slicing, and then converted to category names before visibility filtering.
- Label layout now uses the actual minimum pixel spacing between uneven ticks.
- Added JSDoc and focused tests for domain invariance, reversed axes, datetime/log scales, empty and duplicate values, out-of-domain values, and category normalization.
- Removed the redundant visual snapshot in favor of precise assertions and unit coverage.

`interval` / `pixelInterval` remain inputs to the automatic tick calculation, while `values` takes precedence for rendering.

All CI, visual/performance, preview, and docs checks are green. Please take another look.

PR #685, opened 2026-09-01
feat(legend): add item click event

nikita-jpg:
## Summary

- Adds a public `legend.events.itemClick` callback for discrete legend items.
- Supports both SVG and HTML legends.
- Exposes the item ID, name, current visibility, and series custom data.
- Invokes the callback before the built-in series visibility behavior.
- Allows consumers to cancel the default behavior with `event.preventDefault()`.

## Motivation

Legend items sometimes require application-specific interactions in addition to toggling series visibility. A cancellable click event allows consumers to implement those interactions without replacing the built-in legend renderer.

## API

```ts
const data = {
    legend: {
        events: {
            itemClick: (item, event) => {
                console.log({
                    id: item.id,
                    name: item.name,
                    visible: item.visible,
                    custom: item.custom,
                });

                event.preventDefault();
            },
        },
    },
};
```

The callback receives the legend item data and the native `MouseEvent`.

If `event.preventDefault()` is not called, the existing series visibility behavior remains unchanged. Calling it prevents the built-in visibility change.

## Testing

- `npm run typecheck`
- `npm test -- --runInBand`
- `npm run build`
- Chromium component tests covering the default SVG legend behavior
- Chromium component tests covering cancellation in an HTML legend with `event.preventDefault()`

### DIFF
--- a/src/core/series/types.ts
+++ b/src/core/series/types.ts
@@ -90,7 +90,8 @@ export type PreparedLegendSymbol = (RectLegendSymbol | PathLegendSymbol | Symbol
     bboxWidth: number;
 };
 
-export type PreparedLegend = Required<Omit<ChartLegend, 'title' | 'colorScale'>> & {
+export type PreparedLegend = Required<Omit<ChartLegend, 'title' | 'colorScale' | 'events'>> & {
+    events: ChartLegend['events'];
     hangingOffset: number;
     height: number;
     lineHeight: number;
@@ -120,6 +121,7 @@ export type OnLegendItemClick = (data: {id: string; name: string; metaKey: boole
 export type LegendItem = {
     id: string;
     color: string;
+    custom: MeaningfulAny;
     height: number;
     name: string;
     text: string;
--- a/src/__tests__/components/LegendItemClickTestStory.tsx
+++ b/src/__tests__/components/LegendItemClickTestStory.tsx
@@ -1,0 +1,61 @@
+import React from 'react';
+
+import {ChartTestStory} from '../../../playwright/components/ChartTestStory';
+import type {ChartData} from '../../types';
+
+interface Props {
+    html?: boolean;
+    preventDefault?: boolean;
+}
+
+export const LegendItemClickTestStory = ({html = false, preventDefault = false}: Props) => {
+    const [clickedItem, setClickedItem] = React.useState('');
+    const data = React.useMemo<ChartData>(
+        () => ({
+            legend: {
+                enabled: true,
+                html,
+                events: {
+                    itemClick: (item, event) => {
+                        const custom = item.custom as {key: string};
+                        setClickedItem(`${item.name}:${item.visible}:${custom.key}`);
+
+                        if (preventDefault) {
+                            event.preventDefault();
+                        }
+                    },
+                },
+            },
+            series: {
+                data: [
+                    {
+                        type: 'line',
+                        name: 'First series',
+                        custom: {key: 'first'},
+                        data: [
+                            {x: 0, y: 1},
+                            {x: 1, y: 2},
+                        ],
+                    },
+                    {
+                        type: 'line',
+                        name: 'Second series',
+                        custom: {key: 'second'},
+                        data: [
+                            {x: 0, y: 2},
+                            {x: 1, y: 1},
+                        ],
+                    },
+                ],
+            },
+        }),
+        [html, preventDefault],
+    );
+
+    return (
+        <React.Fragment>
+            <ChartTestStory data={data} />
+            <output data-qa="clicked-legend-item">{clickedItem}</output>
+        </React.Fragment>
+    );
+};
--- a/src/__tests__/legend.visual.test.tsx
+++ b/src/__tests__/legend.visual.test.tsx
@@ -9,6 +9,7 @@ import {ChartTestStory} from '../../playwright/components/ChartTestStory';
 import {groupedLegend, pieHtmlLegendData} from '../__stories__/__data__';
 import type {ChartData, ChartLegend} from '../types';
 
+import {LegendItemClickTestStory} from './components/LegendItemClickTestStory';
 import {LONG_TEXT} from './constants';
 
 const pieOverflowedLegendItemsData: ChartData = {
@@ -127,6 +128,32 @@ test.describe('Legend', () => {
             // When clicking on the legend again, the chart should return to its original state.
             await legendItem.click();
             await expect(component.locator('svg')).toHaveScreenshot();
+        });
+
+        test('Item click event runs before the default SVG legend action', async ({mount}) => {
+            const component = await mount(<LegendItemClickTestStory />);
+            const legendItems = component.locator('.gcharts-legend__item text');
+
+            await legendItems.first().click();
+
+            await expect(component.locator('[data-qa="clicked-legend-item"]')).toHaveText(
+                'First series:true:first',
+            );
+            await expect(legendItems.nth(1)).toHaveClass(/gcharts-legend__item-text_unselected/);
+        });
+
+        test('Item click event can prevent the default HTML legend action', async ({mount}) => {
+            const component = await mount(
+                <LegendItemClickTestStory html={true} preventDefault={true} />,
+            );
+            const legendItems = component.locator('.gcharts-legend__item-text-html');
+
+            await legendItems.first().click();
+
+            await expect(component.locator('[data-qa="clicked-legend-item"]')).toHaveText(
+                'First series:true:first',
+            );
+            await expect(legendItems.nth(1)).toHaveClass(/gcharts-legend__item-text-html_selected/);
         });
 
         const positions = ['top', 'bottom', 'left', 'right'] as const;
--- a/src/components/Legend/index.tsx
+++ b/src/components/Legend/index.tsx
@@ -219,6 +219,28 @@ export const Legend = (props: Props) => {
             svgElement.style('opacity', 0);
 
             const isMac = navigator.platform.toUpperCase().includes('MAC');
+            const handleItemClick = (event: MouseEvent, item: LegendItem) => {
+                legend.events?.itemClick?.(
+                    {
+                        id: item.id,
+                        name: item.name,
+                        visible: Boolean(item.visible),
+                        custom: item.custom,
+                    },
+                    event,
+                );
+
+                if (event.defaultPrevented) {
+                    return;
+                }
+
+                onItemClick({
+                    id: item.id,
+                    name: item.name,
+                    metaKey: isMac ? event.metaKey : event.ctrlKey,
+                });
+                onUpdate?.();
+            };
 
             const htmlElement = select(htmlLayout);
             htmlElement.selectAll('[data-legend]').remove();
@@ -248,13 +270,8 @@ export const Legend = (props: Props) => {
                         .enter()
                         .append('g')
                         .attr('class', b('item'))
-                        .on('click', function (e, d) {
-                            onItemClick({
-                                id: d.id,
-                                name: d.name,
-                                metaKey: isMac ? e.metaKey : e.ctrlKey,
-                            });
-                            onUpdate?.();
+                        .on('click', function (event: MouseEvent, item) {
+                            handleItemClick(event, item);
                         });
 
                     const getXPosition = (i: number) => {
@@ -296,13 +313,8 @@ export const Legend = (props: Props) => {
                                 }
                                 return '0px';
                             })
-                            .on('click', function (e, d) {
-                                onItemClick({
-                                    id: d.id,
-                                    name: d.name,
-                                    metaKey: isMac ? e.metaKey : e.ctrlKey,
-                                });
-                                onUpdate?.();
+                            .on('click', function (event: MouseEvent, item) {
+                                handleItemClick(event, item);
                             })
                             .html((d) => d.text);
                     } else {
--- a/src/core/series/prepare-legend.ts
+++ b/src/core/series/prepare-legend.ts
@@ -86,6 +86,7 @@ export async function getPreparedLegend(args: {
         verticalAlign: get(legend, 'verticalAlign', legendDefaults.verticalAlign),
         justifyContent: get(legend, 'justifyContent', legendDefaults.justifyContent),
         enabled,
+        events: legend?.events,
         hangingOffset: itemHangingOffset,
         height,
         itemDistance: get(legend, 'itemDistance', legendDefaults.itemDistance),
--- a/src/core/types/chart/legend.ts
+++ b/src/core/types/chart/legend.ts
@@ -1,4 +1,17 @@
+import type {MeaningfulAny} from '../misc';
+
 import type {BaseTextStyle} from './base';
+
+export interface ChartLegendItemClickData {
+    id: string;
+    name: string;
+    visible: boolean;
+    custom?: MeaningfulAny;
+}
+
+export interface ChartLegendEvents {
+    itemClick?: (data: ChartLegendItemClickData, event: MouseEvent) => void;
+}
 
 export interface ChartLegendItem {
     enabled?: boolean;
@@ -12,6 +25,7 @@ export interface ChartLegendItem {
 }
 
 export interface ChartLegend extends ChartLegendItem {
+    events?: ChartLegendEvents;
     /**
      * Different types for different color schemes.
      * If the color scheme is continuous, a gradient legend will be drawn.


korvin89:
🤖 **AI generated**

## On the cancellation mechanism

Thanks for picking this up — a legend item click callback is genuinely useful, and I'd like to see it land. My concern is not with the event itself but with `event.preventDefault()` as the way to cancel the built-in behavior. I'd suggest we reshape that part before merging.

A few reasons:

**It mixes layers.** Legend selection is React component state (`selectedLegendItems` in `useChartInnerProps.ts:154`). `preventDefault()` is a browser mechanism for cancelling *browser* default actions — following a link, submitting a form, starting a text selection. Nothing browser-default is being cancelled here; the flag is repurposed to mean "skip my `setState`". The mechanism and the meaning come from different layers, which is what makes it hard to reason about.

**It is invisible to the type system.** Nothing in `(data: ChartLegendItemClickData, event: MouseEvent) => void` tells a reader that mutating the second argument changes library behavior. It's a side channel you can only learn from documentation.

**We don't own the flag.** If anything upstream already called `preventDefault()` on that click — a capturing document listener, an overlay library, a host app handler — legend toggling silently stops working, even though `itemClick` never asked for it.

**There's no precedent for it here, and one pointing the other way.** `chart.events.click` and `pointermove` (`src/core/types/chart/chart.ts:16-17`) are pure notifications, and they're the only callbacks of this kind in the config.

**Prior art doesn't do it this way either.** I looked at how other libraries handle this. Cancellable legend events do exist — Highcharts, Plotly and AG Charts all have one — but none of them overloads the native `MouseEvent`. Highcharts and AG Charts synthesize their own event object with their own `preventDefault`, and Highcharts deliberately nests the DOM event as `browserEvent`; they had to explain in highcharts#8068 that their `preventDefault` is *not* the DOM one. Plotly uses a return value instead. Meanwhile, libraries designed React-first — Recharts, and MUI X Charts, which shipped this exact feature in September 2025 after explicitly benchmarking Highcharts/ECharts/AG Charts — chose declarative visibility props and a non-cancellable notification.

**It's hard to walk back.** Once `preventDefault()` ships, we have to honour it forever. Highcharts 11.4.4 moved its default handler from the series-level to the legend-level event and the deprecated hook's `preventDefault()` became a silent no-op — no type error, no warning, docs still advertising it. A cancellable event's contract is effectively "where we attached our default handler", which is an implementation detail.

**And it only solves half the problem.** `preventDefault()` lets a consumer stop us from changing the selection, but gives them no way to *set* it. In practice, cancelling keeps `selectedLegendItems === null`, so control silently falls back to `series[].visible` — an undocumented round-trip through the whole data object. It's also all-or-nothing: one un-prevented click and the chart owns the selection permanently, with no reset. Who owns visibility ends up depending on click history rather than on configuration.

## What I'd suggest instead

I think the underlying need here isn't "a click event" but "let me own visibility", so I'd answer it with a controlled/uncontrolled pair and keep the event as a pure notification.

```ts
export interface ChartLegendItemClickData {
    /** `series.legend.groupId` of the clicked item. */
    id: string;
    /** Text shown in the legend for this item. */
    name: string;
    /** Ids of all items currently rendered in the legend, in render order. */
    items: string[];
    /** Ids of the items visible at the moment of the click. */
    visibleItems: string[];
}

export interface ChartLegendEvents {
    /**
     * Fired when a legend item is clicked, before the visible set is applied.
     * Informational only — it cannot cancel anything.
     * To take over visibility, use `legend.visibleItems`.
     */
    itemClick?: (data: ChartLegendItemClickData, event: MouseEvent) => void;
}

export interface ChartLegend extends ChartLegendItem {
    events?: ChartLegendEvents;
    /**
     * Ids (`series.legend.groupId`) of the legend items that are currently visible.
     *
     * Omit it and the chart manages visibility itself, seeded from `series.visible`
     * — uncontrolled.
     *
     * Provide it and the chart renders exactly these items and never changes the set
     * on its own; update it from `events.itemClick` — controlled. Takes precedence
     * over `series.visible` and requires an explicit `series.legend.groupId`.
     *
     * Series with `legend.enabled: false` are not part of the legend and stay
     * governed by their own `series.visible`.
     */
    visibleItems?: string[];
}
```

Four payload fields, none derivable from another, none coupled to our default algorithm. The raw `MouseEvent` is still there, so a consumer who wants their own modifier convention can read it directly — which is why I'd leave the normalized `metaKey` internal rather than exposing it.

I'd also keep the computed next selection out of the payload. It would be convenient, but it would make our default algorithm part of the public contract — and then changing that behavior later breaks controlled consumers silently, with no type error and no warning. That's the same trap as `preventDefault`, just wearing a different hat.

**Semantics.** Without `visibleItems`, behavior is exactly what it is today. With it, we render precisely what's passed and never mutate anything ourselves. "Notify me but don't change anything" is then just passing a constant array, so no extra opt-out flag is needed.

## Details

**Identity.** This is the part I'd think hardest about. `series.legend.groupId` is the only consumer-authored identifier in the public API, and it defaults to `getUniqId()` — `gravity-chart.<5 random chars>` — regenerated on every `getPreparedSeries` run. So `id` in the payload is currently unusable for identification unless the consumer sets `groupId` explicitly. I'd make `groupId` required in controlled mode and add dev warnings: one when `visibleItems` is set but some series has no explicit `groupId`, and one when `visibleItems` contains ids matching no legend item. Worth calling out in TSDoc that for pie/funnel the id comes from `data[].legend.groupId`, not from the series.

This is also why I'd drop `custom` from the payload. As it stands, `custom` becomes the de facto answer to "which item was clicked", which turns a field documented as a free-form user subspace into part of our contract — one we can't validate or guarantee. It also doesn't work for pie/funnel: there one legend item is one data point, but `custom: series.custom` (`prepare-pie-series.ts:93`) gives every slice the parent series' value, and the per-point `data[].custom` never reaches the legend. Fixing identity removes the need for it entirely.

**Implementation.** I think this ends up smaller than the current diff:

- `PreparedLegend`, `LegendItem` and `prepare-legend.ts` stay untouched — no `events`, no `custom`, so nothing leaks into `PrepareSeriesArgs.legend` and every series plugin's `prepareSeries`.
- `OnLegendItemClick` gains a second `event: MouseEvent` argument; `<Legend>` stays dumb and just forwards `{id, name, metaKey}` plus the event. That also lets the two listeners become `.on('click', handleItemClick)` directly, since both wrappers already match d3's `(event, datum)` signature.
- Everything else lives in `useChartInnerProps`, next to the existing selection logic: `activeLegendItems = data.legend?.visibleItems ?? selectedLegendItems ?? getActiveLegendItems(...)`, then call `data.legend?.events?.itemClick?.(...)` and apply `setSelectedLegendItems(next)` only when uncontrolled. Reading the handler off the raw config there mirrors how `data.chart?.events?.click` is read in `ChartInner/index.tsx:167-168`.
- One adjustment needed: `onlyItemSelected` (`useChartInnerProps.ts:411`) is computed from `selectedLegendItems`, which stays `null` forever in controlled mode, so the "click the only visible item to restore all" branch would never fire. Computing it from `activeLegendItems` instead is equivalent today except in one case — the first click on a chart configured with a single visible series would restore all rather than isolate. I think that's the more correct behavior, but it is a behavior change worth agreeing on.

**Tests.** The story component is necessary — functions nested inside `data` don't survive Playwright CT prop serialization, which is exactly why the four sibling components in `src/__tests__/components/` exist. But it could be much thinner: Playwright does relay top-level function props back to Node (see `onRender={handleRender}` with `expect.poll` in `line-series.visual.test.tsx:860-874`), so the story could take `onItemClick` as a prop and optionally own `visibleItems`, dropping the `useState`, the `<output data-qa>` element and the `${item.name}:${item.visible}:${custom.key}` string round-trip in favour of asserting on the payload object directly.

I'd cover: uncontrolled (the chart applies the change itself), controlled (the chart follows the prop and changes nothing on its own), and "clicked but the app didn't update its state — nothing moved", each on SVG and HTML. The third one matters: MUI X has a bug in exactly that spot, where the click path toggles internal state without checking whether the component is controlled.

One note on the current tests either way — `Item click event runs before the default SVG legend action` doesn't actually observe ordering. `visible` is read from the d3-bound datum, which doesn't change synchronously, so the assertion would pass equally well if the callback ran after `onItemClick`.

## Separate from this PR

Three things I ran into while reading, all pre-existing:

**Stale selection can blank the chart.** After a legend click, `selectedLegendItems` holds group ids. If the consumer then updates `data`, `getPreparedSeries` re-runs and regenerates every default `groupId`, so the stored selection matches nothing and `getVisibleSeries` hides every legend-enabled series. A remount doesn't save it — `<ChartInner>` is mounted without a data-derived `key`. I haven't run this, it's from reading the code, so it's worth confirming. Same root cause as the identity point above, and there's no coverage for legend + data update in `src/__tests__/update-chart/`.

**The click semantics aren't documented anywhere.** Plain click isolates (and restores all when the clicked item is already the only visible one); ⌘/Ctrl adds or removes a single item. That's in neither the types nor `docs/diplodoc/pages/guides/legend.md`. Since this PR formalizes legend clicks as public API, I think it's the right moment to write it down.

**The "restore all" branch uses the wrong set.** `getAllLegendItems` (`src/core/series/utils.ts:21-23`) is a plain `series.map(s => s.legend.groupId)` — no filter on `legend.enabled`, no dedup by group. So `allItems.length === 1` in `handleLegendItemClick` counts series that aren't in the legend at all, which means "click the only legend item to hide it" won't fire whenever the chart has any legend-disabled series; and the restore-all branch writes duplicate and non-legend ids into the active set. Harmless downstream, since only `includes()` is used, but it makes the branch conditions wrong.

## Smaller notes

- The new public types have no TSDoc, while every other field in `ChartLegend` does. The API reference is generated from TSDoc and gitignored, so the feature would ship with an empty description. The legend guide isn't updated either, and there's no Storybook story or `docs/examples` entry — `chart.events.click` has `action('chart.events.click')` wired into a dozen playground stories.
- `component.getByTestId('clicked-legend-item')` would be more idiomatic than `locator('[data-qa="…"]')`, since `playwright.config.ts:51` sets `testIdAttribute: 'data-qa'`.
- `LegendItemClickTestStory` imports `ChartTestStory` from `../../../playwright/components/…`, while all five sibling components use `ChartStory` from `../../__stories__/ChartStory`. There's also a `~playwright/*` alias if the reverse import is intended.
- If a consumer's handler throws, the exception escapes the d3 listener before the visibility update runs, so the legend silently stops toggling. `chart.events.click` behaves the same way, so it's consistent — but a `try/catch` would be more robust.
- Worth running `npm run test:chart-config`. With `visibleItems` being plain data rather than a callback, `normalizeSchema` won't strip it, so the committed `chart-config.schema.summary.json` snapshot will need regenerating — unlike the current shape, where `events` would be stripped entirely.
- Not introduced here, but adjacent: legend items are still plain `<g>`/`<div>` with `cursor: pointer` and no `role`, `tabindex` or keyboard handling. The PR that makes legend clicks public API feels like a natural moment to make them focusable and Enter/Space operable.

---

@kuzmadom could you take a look at this? The API shape here is a one-way door once released, so I'd rather we agree on it before merging. Let's hold off until you've had a chance to reply.

kuzmadom:
@nikita-jpg could you share one or two concrete use cases for this API? What should happen when a legend item is clicked, and should the default visibility behavior still apply?

  An example that requires preventing the default behavior would be especially helpful. The callback itself seems useful; concrete examples would help us decide whether a cancellable event is enough or whether the application needs to explicitly control which legend items are visible.

nikita-jpg:
One concrete use case is treating a legend item as a navigation action rather than a visibility toggle. For example, clicking the “Created” or “Resolved” legend item opens a page with issues filtered by that series and the chart date range. The chart itself should remain unchanged, so the default visibility behavior must not apply.

Highcharts supports exactly this through `series.events.legendItemClick`: the handler performs the custom action and calls `event.preventDefault()` to keep the series visible.

So the callback is useful, but for this case it also needs a way to cancel the built-in legend action. I agree that this does not necessarily have to reuse the native `MouseEvent.preventDefault()`; a library-owned event or a callback return value could express it instead.

src/core/series/types.ts
```diff
@@ -120,6 +121,7 @@ export type OnLegendItemClick = (data: {id: string; name: string; metaKey: boole
 export type LegendItem = {
     id: string;
     color: string;
+    custom: MeaningfulAny;
```
kuzmadom: Could we clarify which object's `custom` is exposed here? For pie/funnel, the clicked legend item represents a data point, but the current implementation exposes the parent series' `custom`. Grouped items also need an explicit rule.

  I'd prefer exposing series and point context separately, following the distinction used by tooltips, so consumers can access the appropriate metadata without implicit selection or fallback rules. This doesn't necessarily require exposing the full series object.

kuzmadom:
> One concrete use case is treating a legend item as a navigation action rather than a visibility toggle. For example, clicking the “Created” or “Resolved” legend item opens a page with issues filtered by that series and the chart date range. The chart itself should remain unchanged, so the default visibility behavior must not apply.

For the navigation use case, I'd prefer a separate option to disable the built-in legend action while keeping `itemClick` as a notification:

  ```ts
  legend: {
      itemClickAction: 'none',
      events: {
          itemClick: (item, event) => {
              navigateToIssues(item.id, dateRange);
          },
      },
  }
  ```

  `itemClickAction` is a working name. The proposed contract would be:

  - `'default'` (the default): preserve the existing selection behavior, including modifier keys.
  - `'none'`: leave visibility unchanged, but still invoke `itemClick`.
  - The callback's return value and the native event's `defaultPrevented` flag do not control series visibility.

  This makes the interaction policy explicit in the configuration and keeps it independent of the application's event handler. It also covers the described navigation scenario without requiring the application to manage legend state.

kuzmadom (changes requested):

src/core/series/types.ts
nikita-jpg: Removed `custom` from `ChartLegendItemClickData` and from the internal legend click payload in 49ca4d1. The current use case only needs the legend item `id`; series and point context can be designed separately if it is needed later.

korvin89:
🤖 **AI generated**

🔴 blocker — `itemClickAction: 'none'` changes tooltip behavior, and does it differently for SVG and HTML legends

Thanks for the rework — dropping the cancellation, removing `custom`, the TSDoc, the guide section and the widened test matrix all look good to me. One thing I'd ask to fix before this merges.

**What happens.** In `handleItemClick` (`src/components/Legend/index.tsx:222-241`), the `'none'` branch returns early:

```ts
legend.events?.itemClick?.({...}, event);

if (legend.itemClickAction === 'none') {
    return;
}

onItemClick({...});
onUpdate?.();
```

`onUpdate` is `unpinTooltip` (`ChartInner/index.tsx:434`), and it does more than the name suggests (`useChartInnerState.ts:47-50`): besides clearing the pin, it dispatches `HOVER_SHAPE` with `undefined`, which resets the tooltip contents, the shape hover highlight and the crosshair (`useCrosshair/useCrosshairHover.ts:25`). It's the same call used on pointer leave (`useChartInnerHandlers.ts:77, 135`).

With `'none'` that never runs — and at that point the two legend renderers start behaving differently, because of the pinned tooltip's outside-click handler (`Tooltip/index.tsx:71-76`, wired only while pinned at `:90`):

```ts
if (svgContainer?.contains(e?.target as HTMLElement)) {
    return;
}
onOutsideClick?.();   // → unpinTooltip
```

`svgContainer` is the `<svg>` (`ChartInner/index.tsx:444`), while the HTML legend lives in the `html-layer` div, which is a **sibling** of that `<svg>`, not a descendant (`:458`). So with a pinned tooltip and `itemClickAction: 'none'`:

- **SVG legend** — the click target is inside the `<svg>`, the handler early-returns, and the tooltip stays pinned with its hover highlight and crosshair intact.
- **HTML legend** (`legend.html: true`) — the click target is outside the `<svg>`, so it counts as an outside click, `onOutsideClick` fires, and the tooltip unpins.

**Repro.** Pin a tooltip, then click a legend item with `itemClickAction: 'none'` — once with `legend.html: false`, once with `true`.

**Suggested fix.** Hoist `onUpdate?.()` above the `itemClickAction` check, so the pin and hover reset always happen and both renderers stay in step — `itemClickAction` then affects exactly what it promises:

```ts
legend.events?.itemClick?.({...}, event);
onUpdate?.();

if (legend.itemClickAction === 'none') {
    return;
}

onItemClick({...});
```

If keeping the tooltip pinned under `'none'` is deliberate, then the HTML path needs aligning too and the behavior should be spelled out in the TSDoc and the guide — but I'd prefer the first option. Either way it would be worth a test with a pinned tooltip covering both renderers, since the existing four cases don't touch tooltip state.

One caveat: this comes from reading the code, including the assumption that uikit's `Popup` `onOpenChange` fires on an outside click. I haven't run it, so please sanity-check the repro before acting on it.

PR #615, opened 2026-05-19
feat: declarative tooltip row cells and per-series valueFormat

kuzmadom:

PR #687, opened 2026-09-03
fix(tooltip): give a custom rowRenderer the series color for area and line

### LINKED ISSUE #686: Area tooltip rows lose the series color since 1.58
## Summary

Since `@gravity-ui/charts` 1.58, tooltip rows for **area** series no longer show the series color. The colored swatch that used to be painted for every row is rendered empty; the rest of the row (name, value, percentage) is unchanged.

We observe it through a custom `tooltip.rowRenderer` that paints a swatch from the `color` argument it receives:

```tsx
const rowRenderer: ChartTooltip['rowRenderer'] = ({name, formattedValue, color, className}) => (
    <tr className={className}>
        <td><div style={{backgroundColor: color}} /></td>
        <td>{name}</td>
        <td>{formattedValue}</td>
    </tr>
);
```

## Versions

| | |
|---|---|
| Last known good | **1.57.3** |
| Reproduced on | **1.59.0** |
| Consumed via | `@gravity-ui/chartkit` 8.2.0 |
| React | 19 |

The change that looks responsible landed in 1.58.0 (see below), so the first affected release is most likely 1.58.0 — we bisected only between 1.57.3 and 1.59.0.

## Steps to reproduce

1. Render a stacked percent ("normalised") area chart with three series.
2. Provide a custom `tooltip.rowRenderer` that paints a swatch using the `color` argument, as above.
3. Hover a point so the tooltip appears.

**Expected:** every tooltip row shows a swatch filled with its series color.
**Actual:** the swatch cell stays empty.

No per-point colors are set in the series data (points only carry `x`, `y`, `label`, `custom`), and no `gradient` is configured on the series. Series-level `color` is set.

## How the cause was isolated

Hit while upgrading a large application. We held `@gravity-ui/charts` at 1.57.3 with an override while every other dependency in the upgrade stayed identical, and re-ran the same screenshot tests:

- charts **1.59.0** — area tooltip screenshot tests fail, swatches missing
- charts **1.57.3** — the same tests pass, swatches present

Nothing else changed between the two runs, so the regression is in charts rather than in the surrounding stack.

## Where it appears to come from

`src/core/shapes/area/get-tooltip-data.ts` now sets an explicit `color` on every chunk:

```ts
acc.push({
    color: getMarkerFill(p, p.series.color),
    data: p.data,
    series: p.series as AreaSeries,
    ...
});
```

`DefaultTooltipContent` resolves a cell value as:

```ts
get(chunk, cell.source)
    ?? get(chunk, 'data.' + cell.source)
    ?? get(chunk, 'series.' + cell.source)
```

For `cell.source === 'color'` this used to fall through to `chunk.series.color`, because area chunks carried no own `color`. Now the first lookup wins and the fallback is never reached.

One honest gap: reading the source alone we could **not** explain why the resulting value ends up unusable. With neither `fill` nor `color` on the point — which is our case — `getMarkerFill(p, p.series.color)` should return `p.series.color`, i.e. exactly what the old fallback produced. Something in that chain evidently resolves differently at runtime, and that is the part worth checking first.

Two more pointers:

- the same `color: getMarkerFill(...)` line was added to `src/core/shapes/line/get-tooltip-data.ts`, so line series are likely affected as well (not verified);
- `src/core/series/prepare-legend.ts` also changed within the same version range and has not been ruled out.

## Screenshots

Tooltip of a normalised area chart, same chart and data in both cases.

| charts 1.57.3 (expected) | charts 1.59.0 (actual) | diff |
|---|---|---|
|Image|Image|Image|

korvin89:
🤖 AI generated

Fixes #686

## What was happening

Since 1.58, a custom `tooltip.rowRenderer` on an area series received SVG swatch markup in `color` instead of a color, so `backgroundColor: color` resolved to nothing and the cell stayed empty.

The cause is not the one the issue points at. `getMarkerFill(p, p.series.color)` in `get-tooltip-data.ts` resolves correctly. What broke is elsewhere in the same PR (#669): the area `color` cell moved from the declarative `source: 'color'` + `format` form to a **function `source`** that returns the rendered swatch.

And `DefaultTooltipContent` applies `format` only when it renders the default row — in the `rowRenderer` branch the format is never applied. So a string `source` yields the raw value, while a function `source` yields the rendered markup:

| series | what reached `rowRenderer({color})` |
|---|---|
| `area` | `<svg height="8" width="16">…fill="#ff0000"…</svg>` |
| `line` | `<svg height="8" width="16">…stroke="#ff0000"…</svg>` |
| `bar-x` | `#ff0000` |

`line` has the same bug, introduced earlier by the same architectural move. Before #615 (1.54.0) `DefaultTooltipContent` resolved the color separately from the cells — `get(data, 'color') || get(series, 'color')` — so every series type reached `rowRenderer` with a real color, and the swatch markup was a separate concern. #615 collapsed the value and its rendering into `source` and broke `line`; #669 did the same to `area`.

## What this changes

Restores the split every other plugin already follows: **`source` resolves the value, `format` renders the swatch.**

- **`area`** — the cell goes back to the declarative form it had before 1.58, matching the other ten plugins.
- **`line`** — `rows` becomes a per-chunk function (the existing `SeriesPlugin['tooltip']['rows']` mechanism, already used by waterfall) so the formatter can close over the `dashStyle` / `lineWidth` its line symbol needs.

## Out of scope

`waterfall` passes `color: null` to `rowRenderer` because its tooltip rows have no `color` cell at all. That predates this issue and looks deliberate — waterfall bars are colored by rise / fall / total, so a single series color would be misleading.


src/core/types/chart/tooltip.ts
```diff
@@ -161,6 +161,15 @@ export type ChartTooltipRowRendererArgs = {
...
+    /**
+     * Color of the hovered point as a plain CSS color value — never the built-in swatch
+     * markup, since a custom row paints its own. Resolved from the point's own color when
+     * it has one, and from the series color otherwise; for a gradient series it is the
+     * gradient sampled at that point.
+     *
+     * Not provided for series types whose tooltip has no color cell (e.g. `waterfall`).
+     * When `rows[].cells` overrides the color cell, its `source` decides the value instead.
+     */
     color?: string;
```
dvortsovs, line 173: Align the new `color` contract with the runtime behavior.

The JSDoc states that no value is passed for series without a color cell.

Accordingly, the type only allows `string | undefined`. However, the resolver returns `null` and then explicitly passes it to `rowRenderer`.

This occurs, for example, with waterfall charts. We should either normalize a missing color to `undefined` or change the type to `color?: string | null` and document the meaning of `null`.

korvin89: 🤖 AI generated

Good catch — fixed in 71b6699cb by normalizing to `undefined`.

Went with that over widening the type to `string | null`: the arg type is handed to the consumer, so widening it would break anyone who typed their renderer's `color` as `string | undefined`. Normalizing keeps the public API untouched and makes the runtime match what the JSDoc already claimed.

`getTooltipRowCellValue` still returns `null` for a missing cell — that is shared with `name` and `value`, and `value` is typed `string | number | null | undefined`, so it is legitimate there. Only the `color` argument is normalized at the call site.

Added a regression test using a `waterfall` chunk, since that is the type with no color cell; it fails without the change.

Worth flagging separately: `name` is typed `string` but gets the same `null` when a custom `rows[].cells` omits a name cell. That predates this PR and I left it alone — happy to fix it here if you would rather have both.

dvortsovs (approved):

PR #692, opened 2026-09-17
fix: yield between chart resize preparation phases

### LINKED ISSUE #691: Chart resize updates can block concurrent interactions
## Problem

`Chart` handles `ResizeObserver` notifications with a trailing 200 ms debounce. During continuous container resizing, updates can be postponed until notifications stop. The resulting React state update is also scheduled as urgent work and can contend with interactive updates.

## Expected behavior

- Resize state updates should run as transitions where supported.
- Continuous resize should still update at least every 200 ms.
- React 17 compatibility should be preserved.

daniil-filippov-dev:
🤖 AI generated

Closes #691.

Keep periodic chart updates during continuous container resize without running all preparation in one uninterrupted task. Yield to the browser before axes and shape preparation, discard obsolete work after a newer effect or unmount, and remove the ineffective transition wrapper. Y-axis tick generation also avoids collecting a full domain when it is unused.

The performance test now requires interaction p95 below 50 ms and at least five actual SVG path changes before resizing stops. Explicit `reflow()` remains debounced; `reflow({immediate: true})` still updates dimensions immediately.

In `mcr.microsoft.com/playwright:v1.56.1-jammy`, `chromium-perf`, one worker: the previous PR head measured 54.9–62.4 ms interaction p95 (3 runs); this revision measured 32.0–43.0 ms (4 runs), with 7 live path changes per run. Main measured 1.0–1.2 ms locally but rendered no intermediate path changes. These results show reduced blocking compared with the previous live-resize implementation, not zero overhead compared with deferring all chart work.

Validation: 389 unit tests, typecheck, scoped ESLint, package build, and four performance runs passed.


src/components/index.tsx
```diff
@@ -78,7 +78,22 @@ export const Chart = React.forwardRef<ChartRef, ChartProps>(function Chart(props
...
-        debounced.current = debounce(handleResize, 200);
+        const startTransition = (
+            React as typeof React & {
+                startTransition?: (callback: () => void) => void;
+            }
+        ).startTransition;
+        debounced.current = debounce(
+            (options) => {
+                if (typeof startTransition === 'function') {
+                    startTransition(() => handleResize(options));
```
korvin89: 🤖 **AI generated**

🟠 **should-fix**

It looks like `startTransition` here only covers the lightweight part of the update: `setDimensions` in `Chart` and the re-render of `ChartInner` with the new `width`/`height`. The expensive work (preparing series, axes and shapes) starts later, in the async `useEffect` in `src/components/ChartInner/useChartInnerProps.ts:159`, and is committed via `setState(newStateValue)` at line 373.

Updates scheduled from an effect or after an `await` are not part of the transition. So the heavy re-render still runs at default priority. I checked this in a local instrumented copy: `ReactCurrentBatchConfig.transition` is set when `setDimensions` is called and is `null` when `setState(newStateValue)` runs.

Because of that, I'm not sure the goal from #691 (resize updates not blocking interactions) is actually reached. Could you please measure it? For example, drag-resize a large chart while interacting with it, before and after this change, and share the results.

src/components/index.tsx
```diff
@@ -78,7 +78,22 @@ export const Chart = React.forwardRef<ChartRef, ChartProps>(function Chart(props
...
+                } else {
+                    handleResize(options);
+                }
+            },
+            200,
+            {maxWait: 200},
```
korvin89: 🤖 **AI generated**

🟠 **should-fix**

With `maxWait: 200`, a continuous resize now recalculates the whole chart about 5 times per second, instead of once at the end. That recalculation isn't deferred (see the comment on line 89), so on heavy charts it may add main-thread load while dragging. That's the opposite of what #691 is after.

Consumers will also notice a change: `onResize` and `onReady` (`src/components/ChartInner/index.tsx:324-328`) now fire every ~200 ms during a drag. With resize notifications every 16 ms for 1 s, I got 5 `onResize` calls.

Could you check this trade-off in the same measurement? If it's intended, it would be great to mention the new callback frequency in the commit message, so it shows up in the changelog.

korvin89 (commented):
🤖 **AI generated**

**Severity legend**

src/components/index.tsx
```diff
@@ -78,7 +78,22 @@ export const Chart = React.forwardRef<ChartRef, ChartProps>(function Chart(props
 
     const debuncedHandleResize = React.useMemo(() => {
         debounced.current?.cancel();
-        debounced.current = debounce(handleResize, 200);
+        const startTransition = (
+            React as typeof React & {
+                startTransition?: (callback: () => void) => void;
+            }
+        ).startTransition;
+        debounced.current = debounce(
```
korvin89: 🤖 **AI generated**

🟠 **should-fix**

The public `ChartRef.reflow()` (line 107) uses the same wrapper. So it now also runs as a transition and follows the `maxWait` behavior, even though #691 is only about `ResizeObserver`.

`reflow()` passes `{force: true}`, so every call creates a new `dimensions` object. If a consumer calls `reflow()` on every frame (e.g. during a sidebar CSS animation), `onResize` will now fire every 200 ms even when the size doesn't change. Before, those calls were merged into one.

If that isn't intended, maybe we could keep the plain debounce for `reflow()` and use the transition and `maxWait` only in the `ResizeObserver` callback?

src/components/index.tsx
```diff
@@ -78,7 +78,22 @@ export const Chart = React.forwardRef<ChartRef, ChartProps>(function Chart(props
...
+            (options) => {
+                if (typeof startTransition === 'function') {
+                    startTransition(() => handleResize(options));
+                } else {
+                    handleResize(options);
+                }
+            },
+            200,
```
korvin89: 🤖 **AI generated**

🟠 **should-fix**

`debounce(fn, 200, {maxWait: 200})` is exactly lodash's `throttle(fn, 200, {leading: false})`: lodash implements `throttle` as `debounce` with `maxWait` equal to `wait`. After this change, the names `debounced` and `debuncedHandleResize` no longer describe what the function does, and the two `200` values could drift apart later.

What do you think about switching to `throttle(..., 200, {leading: false})`, or at least adding a named constant with a short comment? The `startTransition` lookup with the type cast could also move into a small module-level helper, to keep the `useMemo` body simple.

src/components/__tests__/chart-reflow.test.tsx
```diff
@@ -133,6 +133,41 @@ describe('Chart/reflow', () => {
         });
     });
 
+    test('continuous container resize triggers onResize after maxWait', async () => {
```
korvin89: 🤖 **AI generated**

🟠 **should-fix**

The unit test covers `maxWait`, but the main point of this PR is performance: resize updates shouldn't block interactions. Could you please add a performance test for this case? The existing `@perf` tests (e.g. in `src/__tests__/line-series.visual.test.tsx`) could be a good template. For example, keep resizing the container of a large chart while interacting with it. That would give us a measurement for this change and also guard against future regressions.

src/components/index.tsx
korvin89: 🤖 **AI generated**

🔴 **blocker**

Thanks for adding `runInTransition` and the perf story — that gave us the measurement. I ran the new `@perf` test in the Playwright docker image (`mcr.microsoft.com/playwright:v1.56.1-jammy`, project `chromium-perf`, `--workers=1`), 4 repeats per variant:

| variant | interactionP95 | interactionCount | resizeCount |
| --- | --- | --- | --- |
| `main` (before this PR) + the new test | 1.6–1.7 ms | 31 | 1 |
| this PR (b2d5738) | 104–124 ms | 16–18 | 7–8 |
| this PR with `runInTransition` removed | 102–124 ms | 16–18 | 8 |

Two things stand out:

- The transition doesn't change anything measurable: the numbers with and without `runInTransition` are the same. The dominant cost is the synchronous preparation work (series, axes, shapes) inside the effect in `useChartInnerProps`, which blocks the main thread regardless of the update's priority.
- Compared to `main`, interaction latency during a continuous resize goes from ~1.6 ms to ~120 ms, and only about half of the interactions get their own render (31 → ~17).

So the periodic updates from #691 are delivered, but the other half of the issue — resize updates not contending with interactions — measures worse than before this PR. Could we look at the approach again before merging? Happy to re-run the measurement on anything you try.

src/components/__tests__/chart-reflow.test.tsx
korvin89: 🤖 **AI generated**

🟠 **should-fix**

Thanks for adding the perf test. The problem is that as written it doesn't guard what it was added for: it passes both with and without `runInTransition`, and the only assertion that fails on pre-PR code is `resizeCount >= 5`. So right now it locks in the periodic updates, not the responsiveness (numbers in my comment on `src/components/index.tsx`).

The `interactionP95 < 200` budget sits far above both the pre-PR value (~1.6 ms) and the current one (~120 ms), and `interactionCount >= 10` doesn't catch the drop from 31 interactions to ~17. Could you tighten both thresholds so that a regression in interaction latency actually fails the test? It would also be great to align it with the existing `@perf` tests, which take a median over 10 runs instead of measuring a single pass.

korvin89 (approved):

daniil-filippov-dev:
🤖 AI generated

**History note**

This PR started as a resize performance improvement using `startTransition` and periodic updates every 200 ms. Measurements showed that `startTransition` did not cover expensive chart preparation, while live resize increased interaction latency compared with `main`. After rework, the PR solves a different problem: it renders intermediate resize states and yields to the event loop between expensive preparation phases.

**Resolution:** the PR was moved back to draft. Before continuing, we need to confirm that live resize is desired product behavior. Without that requirement, the existing trailing debounce is simpler and faster.

